PR #63, opened 2021-04-28
Object detection 2d

jelledouwe:
Dear reviewers,

This branch contains the learner class: DetrLearner, which is a learner that can be used for 2d object detection. DETR is an algorithm that was developed by Facebook Research (https://arxiv.org/abs/2005.12872). The original license type is Apache 2.0. The code is based on code that is available here: https://github.com/facebookresearch/detr. Since I had to modify the code for the required functionalities of the learner, the algorithm code is inside the learner module, rather than linking to the original repo. I took care of license procedures to the best of my knowledge, but I must say that I am little experienced regarding this. Next to adding the learner module, I also added a unit test script and a reference doc.

Kind regards,

Jelle


src/opendr/perception/object_detection_2d/detr/detr_learner.py
```diff
@@ -0,0 +1,932 @@
...
+        Method for resetting the state of the model. Since the model does not
+        have a state, this method simply passes.
+
+        Returns
+        -------
+        None.
+
+        """
+
+    def download_model(
```
passalis, line 661: Even though the signature of this function has not been specified in the first version of specifications, I think it would be nice if we could stick to an existing way of implementing this (e.g., https://github.com/tasostefas/opendr_internal/blob/master/src/opendr/perception/pose_estimation/lightweight_open_pose/lightweight_open_pose_learner.py#L906 ) in order to avoid having to update this later. So, it would be very nice if you could rename and add any arguments (if needed) to follow the function signature provided by the rest of the implementations.

src/opendr/perception/object_detection_2d/detr/detr_learner.py
```diff
@@ -0,0 +1,932 @@
...
+
+            if self.args.distributed:
+                self.model = torch.nn.parallel.DistributedDataParallel(self.model, device_ids=[self.args.gpu])
+                self.model_without_ddp = self.model.module
+
+            self.n_parameters = sum(p.numel() for p in self.model.parameters() if p.requires_grad)
+        else:
+            raise ValueError(self.backbone + " not a valid backbone. Supported backbones:" + str(supportedBackbones))
+    
+    def download_nano_coco(self):
```
passalis, line 749: This function can be provided as part of the generic *download()* method (see the previous comment).

passalis (commented):
Thanks for the contribution! 

Is this tool ready for review? In any case, I have made a few comments on the code regarding this implementations. Also,  one suggestion regarding the target modification would be to have a new target class, e.g,. *COCOBoundingBox*, that would provide the extra functionality. Perhaps this would be a good approach in general in order avoid polluting the generic classes with extra functionality that is specific to specific implementations. In this way, *COCOBoundingBox* would inherit *BoundingBox*, so instances of *COCOBoundingBox* would be directly usable by the user without having to know any internals of the method.

Please note that testing is now available again. So, when you are ready, you can use the "test sources" label to run the style checks and "test tools" to run the tool tests.

src/opendr/perception/object_detection_2d/detr/detr_learner.py
jelledouwe: Thank you Nikolaos for reviewing the code! It is indeed ready for review. I will rename the code and add the arguments to make it more consistent with the other implementations. Also, I will add the COCOBoundingBox target and work on the style checks.

Issue #75, opened 2021-05-20
Test slowdown


LukasHedegaard:
There has been a significant reduction in test speeds over the past few weeks. Thanks to @iliiliiliili and @negarhdr for noticing this.
Can this be alleviated or is it a result of the additional method integrations?

LukasHedegaard:
Fixed in PR #81

PR #81, opened 2021-06-04
Test speed up activity recognition

LukasHedegaard:
This PR speeds up the Human Activity Recognition tests (see #75 )

### Added
- Kinetics downloader skips download if it already exists
- Kinetics micro download (3 videos, 3 classes)

### Updated
- test_x3d_learner and test_cox3d_learner don't delete temp folder. This was done since they share datasets and weights (we thus avoid double download). It is assumed that the CI system deletes the project folders anyways after running tests.


passalis (approved):
Thanks for making tests run faster! You can also report the observed speedup (just for reference).

stefaniapedrazzi (approved):
Thank you!

On my local system the `unittest` system now reports that it takes just 39s (previously 130s).

PR #82, opened 2021-06-07
End to end planning

halil93ibrahim:
Hi All,

This is an initial version of our method on end-to-end local planning.
It's not completely ready to be merged yet, but should already include all the main parts.

- It implements the LearnerRL interface
- Formatted according to PEP-8
- Includes a first version of the documentation

Remaining todo's for myself:
- Tests for code


passalis:
@halil93ibrahim testing is now available again. So, when you are ready, you can use the "test sources" label to run the style checks and "test tools" to run the tool tests.
Please let us know when this PR is ready for review.

stefaniapedrazzi:
I convert this PR to draft.
Please change the status when it is ready for review.

passalis:
Is this PR ready for review? I also see that some tests are failing due to pep8 errors.

ad-daniel:
To summarize what was said today:
- this tool will be integrated as part of a nightly build
- all the images, proto, etc (basically anything inside ardupilot) should be zipped, and uploaded to the FTP server. Instructions should instead be provided as to where to download it from and where to extract it to.
- Since webots and ardupilot are only needed for the demo and not the unittest, a more detailed guide on where to download them and how to install these 2 tools should be provided.
- the unittest for the learner doesn't require webots, so it should run just fine on its own. For it to automatically be launched in the CI, it just needs to be mentioned [here](https://github.com/opendr-eu/opendr/blob/07dee3b59d3487b9c5a93d6946317178a02c9890/.github/workflows/tests_sources.yml#L62) as done for other tools
- I've quickly tested locally to run the unittest and I wasn't working and I think it's due to a conflict with other tools. It seems that the learner or the test itself relies on `tf.contrib` which is no longer available for tensorflow 2.x but only tensorflow 1.x which might be in conflict with other tools. I don't remember the details exactly but the best way to see it is either enabling the test in the CI and look at the logs if/when it fails. Alternatively you can install the toolkit locally and afterwards attempt to run your unittest locally with `python3 -m unittest tests/sources/tools/...` and check if any dependencies clashes occur (by installing the toolkit beforehand, and your dependencies afterwards (which btw are also not currently provided, you can check how it's done for other tools)

halil93ibrahim:
Hi, thanks for the summary.

- I will move the required files to the FTP server today.
- The unit tests for this tool pass on my local computer. For the unit tests of eval and fit functionalities, they are interacting with the provided gym environment, which is in communication with Webots and Ardupilot. I am not sure how I could test them without interaction with the gym environment.
- About the TensorFlow version: the tool depends on the stable-baselines package, which requires TensorFlow 1.14. I will check the conflicts as you explained; however, now I have another issue I cannot solve about installing dependencies (I am explaining in a new bullet).
- In the system, I am getting ModuleNotFoundError for the "gym" package, although I list it in the dependencies file. I could not solve this issue. To check if the other dependencies are installed or not: the packages "stable-baselines" or "tensorflow" is not giving the same error when I list them in the dependencies file. What could be the reason behind this?
- I am ignoring test-tools for other tools in order to get the test results quicker from CI. Is this ok to do?

halil93ibrahim:
Hi @ad-daniel,

I cannot manage to upload the files to the FTP server. Can you help me to upload the following files into the related folder? https://cicloud.csd.auth.gr/owncloud/f/4456338

What do you suggest about the conflict between tf and python versions?

passalis:
Hi @halil93ibrahim , @Pavlos-Tosidis  will upload the files from owncloud to the FTP server.

Pavlos-Tosidis:
> Hi @ad-daniel,
> 
> I cannot manage to upload the files to the FTP server. Can you help me to upload the following files into the related folder? got the link, hid it from here.
> 
> What do you suggest about the conflict between tf and python versions?

Files are uploaded

halil93ibrahim:
> Files are uploaded

Thanks!

ad-daniel:
>   * The unit tests for this tool pass on my local computer. For the unit tests of eval and fit functionalities, they are interacting with the provided gym environment, which is in communication with Webots and Ardupilot. I am not sure how I could test them without interaction with the gym environment.

For the webots side of things yes, you can run it in the CI but since it's a headless system (no screen and no GPU) keep in mind that if you rely heavily on lidars/radars/cameras it will be very slow. So if it's a very complicated simulation it will take a lot of time. Being headless, you need to run Webots in a virtual X, which you can do using `xvfb`. To give you an idea, on the Webots side we do it like [this](https://github.com/cyberbotics/webots/blob/5389e9d01de62f65ef7ed30f4759c375882663fb/.github/workflows/test_suite_linux.yml#L114-L118). I can help you set it up, for now comment out those tests that rely on it if you want to ensure the rest works before jumping into this issue. 

As for Ardupilot, I have no idea, but we can try to have it running as well.

>    * I am ignoring test-tools for other tools in order to get the test results quicker from CI. Is this ok to do?

It's won't really be quicker since the other tests are run in parallel. You can do whatever you want while the PR is in draft, of course it doesn't mean it can be merged in that state after. But if you just want to see if you can reproduce what you have locally then by all means you can.

>    * There is an error caused when TensorFlow 1.14 is tried to be installed. As far as I understand, this is due to python 3.8 supporting TensorFlow 2.2 or later. The longer solution could be to transfer my codes from stable-baselines to stable-baselines3. However, can I test it with python 3.6 in CI?

Yes you can, you can change this line [here](https://github.com/opendr-eu/opendr/blob/e54b82c50d03f64b2415d157f172d25ca409b037/.github/workflows/tests_sources.yml#L42), assuming doing so doesn't break some other dependencies required by other tools, but you can try. However as you said, ultimately this tool needs to play nicely with the others so once you get it working for 3.6, we still might need to find a solution so that there are no conflicts when using Python 3.8 and that the dependencies you rely upon don't conflict with the others.

halil93ibrahim:
Hi,
In order to run tests without simulation running, the gym environment functions are returning randomly with a ROS warning when it is not connected to Ardupilot.
The tool now depends on stable-baselines3 instead of stable-baselines, so there is no conflict between python 3.8 and dependencies.
The tests have passed in the previous attempt, but now in the recent commits, it gives an error related to git. I suppose this is related to the system itself, and I will retry running them tomorrow.

halil93ibrahim:
Hi,
Can you help me to understand the following errors, did my changes somehow cause this?:

  Error: error: 5976 bytes of body are still expected
  fetch-pack: unexpected disconnect while reading sideband packet
  Error: fatal: early EOF
  Error: fatal: unpack-objects failed
  Error: fatal: Fetched in submodule path 'projects/control/eagerx/eagerx', but it did not contain 1badb04121127b8f4988bc76186cb555549aaa88. Direct fetching of that commit failed.
  fatal: 
  Error: The process '/usr/bin/git' failed with exit code 128

ad-daniel:
No, we have the same issue on a different PR, not sure why now all of a sudden. I'll look into it this afternoon

ad-daniel:
Github actions are having issues: https://www.githubstatus.com/, we might need to wait until they are resolved on their end

halil93ibrahim:
Hi, it seems the issue has been resolved, and the tests have passed.

ad-daniel:
I've added webots to the CI in case you want to try to run the full test

halil93ibrahim:
I think it is identical for the learner's side regarding the format of actions and observations.
But if you think we need to run a full test, the environment is required ArduPilot to be installed and built. Then, some rosnodes should be launched before the test. How could this be done?

ad-daniel:
It's up to you really, if you deem that the functionality of the learner class can be ensured without having to use the full pipeline that's enough for me. The purpose of the test is to alert us if a change somewhere else ends up affecting your tool in an undesired way, if that can be done with "random" data all the better. 
Installing webots, ardupilot, and simulating on the cpu certainly will result in a longer time needed to run the test suite (compared to other tools), but if needed it's no problem either.

Perhaps there exist a middle ground between the two. Is it possible to run the full test on your local machine, register the relevant data from webots/ardupilot to a file which we can simply feed it in the context of the CI test? So to use simulation data but without running the actual simulation.

ad-daniel:
@halil93ibrahim Hi again, just so we're clear, what do you intend to do? If you think the PR is fine as it is we can proceed to do the review otherwise I'll put it back on draft

Issue #62, opened 2021-04-28
Tests are printing many warnings


stefaniapedrazzi:
The existing tests currently print lot of warning (cuda, UserWarning, DeprecationWarning, ResourceWarning) that make the logs hard to read.

We should try to fix all this warnings.
In case this won't be possible, the warnings should be filtered and not printed on the console when running the tests.

All the tests seems to print some warnings:
@Pavlos-Tosidis , @iliiliiliili , @jovaroo could you please double check your tests and fix or filter the warnings?

passalis:
Tests have been revised to this end. We can reopen this issue if we observe excessive number of warning being printed out.

PR #88, opened 2021-06-08
Fix EdgeSpeechNets target class and test

jovaroo:
Related to issue #62, fixes the tests printing unnecessary warnings.

Also fixes the implementation to use the Category target class instead of a special SpeechCommand, and removes the unnecessary code created for that earlier.


passalis (commented):
Thanks for this PR. Please also remove the corresponding documentation from *engine-target.md*, since SpeechCommand is no longer used.

jovaroo:
@passalis I have now removed the unnecessary documentation, and the PR is otherwise ready!

passalis (approved):
Thanks! Seems fine to me.

stefaniapedrazzi (approved):
Thank you.

However, it seems that currently the `TEST Edge Speech Nets Learner` is not printing anything at all. This could be misleading for debugging it later because it is not clear if the test has run successfully, it failed or didn't run at all.

PR #89, opened 2021-06-09
C API for Face Recognition

passalis:
This PR will provide a C API for Recognition following the specification defined in D2.1.

Status:

- [x] ONNX inference in C for face recognition
- [x] Database building and feature matching for face recognition
- [x] Documentation for C API, Installation and compilation instructions

The API is located under the projects directory to avoid polluting root src with C code. However, it can be easily moved to the most appropriate location if needed. 

Significant speedup can be obtained using the C API (CPU inference):

| Mode      | FPS |
| ---------------------------| ----------- |
| Python API (PyTorch backend)   | 5 FPS    |
| Python API (ONNX backend)      | 13 FPS  |
| C API (ONNX backend)      | 29 FPS  |


projects/c_api/Readme.md
```diff
@@ -0,0 +1,49 @@
...
+Before using the C API you should have the appropriate version of ONNX runtime installed. 
+You can use the provided script (`install_deps.sh`) to install it, along with a number of other required dependencies.
+Alternatively, you can also use the supplied Makefile that takes care both of the installation and compilation.
+Please note that only Ubuntu 20.04 is currently officially supported.
+
+In order to build the library you should:
+
+0. Install the required binary dependencies:
+
+```make install_dependencies```
```
stefaniapedrazzi, line 18:
```suggestion
```sh
make install_dependencies
```
```

File: projects/c_api/Readme.md [hunk 2]
```diff
@@ -0,0 +1,49 @@
...
+
+1. Build the static library
+
+```make liboperdr```
```
stefaniapedrazzi, line 22:
```suggestion
```sh
make liboperdr
```
```

File: projects/c_api/Readme.md [hunk 3]
```diff
@@ -0,0 +1,49 @@
...
+
+2. (Optionally) Download the necessary resources and verify that everything works as expected:
+
+```
```
stefaniapedrazzi, line 26:
```suggestion
```sh
```

projects/c_api/Readme.md
```diff
@@ -0,0 +1,49 @@
...
+make download
+make runtests
+```
+After building the API you will find the static library (`libopendr.a`) under the `lib` folder.
```
stefaniapedrazzi, line 30: Isn't better to create a dynamic/shared library instead of a static library?

projects/c_api/Readme.md
```diff
@@ -0,0 +1,49 @@
...
+```
+gcc you_program.c -o your_binary -I./include -L./lib -lopendr
+```
+
+## Demo Applications
+C API comes with a number of demo applications inside the `samples` folder that demonstrate the usage of the API.
+You can build them and try them (check the `build` folder for the executables) using the following command:
+```
+makefile demos
+```
```
stefaniapedrazzi, line 44: This should probably be:
```suggestion
```sh
make demos
```
```

File: projects/c_api/Readme.md [hunk 6]
```diff
@@ -0,0 +1,49 @@
...
+Make sure that you have downloaded the necessary resources before running the demo (`makefile download`) and that you execute the binaries from the root folder of the C API.
```
stefaniapedrazzi, line 45:
```suggestion
Make sure that you have downloaded the necessary resources before running the demo (`make download`) and that you execute the binaries from the root folder of the C API. 
```

stefaniapedrazzi (changes requested):
Thank you!
Sorry for the late review!

I just started looking at this PR and globally seems fine.
But I would not put the C API in the projects folder but rather create a dedicated folder in the `src` for sources and in `include` (https://github.com/tasostefas/opendr_internal/tree/master/include) for headers and the tests should be part of the `tests` folder run by the CI.
However, it is fine to keep the `projects/c_api/samples` folder as is.

We should also reduce the logs printed by the Makefile when installing the dependencies and cloning.
By the way, everytime I type `make libopendr` the dependencies are reinstalled and the repositories and resources are downloade. This should be avoided and actions should only be executed when required and when something changed.

I'm usually working on Ubuntu 18.04, so I'm having issues with libboost-filesystem.
On Wednesday I will test on Ubuntu 20.04.
I will also add style checks for C/C++ that are currently missing.

stefaniapedrazzi (commented):
On Ubuntu 20.04 I'm currently getting this error:
```
Downloading resources for C API...
mv: cannot move 'opendrdata.csd.auth.gr/perception/face_recognition/test_data/images' to 'data/database/images': Directory not empty
```

projects/c_api/Readme.md
passalis: Switched into using a shared library.

Makefile
```diff
@@ -44,6 +44,14 @@ unittest:
 	@+echo "Performing unit tests"
 	@+python3 -m unittest discover -s tests/sources/tools/
 
+libopendr:
```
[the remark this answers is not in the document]
passalis: No, it can be compiled as it is. However, I can add the onnx into compilation time dependencies instead of separately compiling it here.

passalis:
> On Ubuntu 20.04 I'm currently getting this error:
> 
> ```
> Downloading resources for C API...
> mv: cannot move 'opendrdata.csd.auth.gr/perception/face_recognition/test_data/images' to 'data/database/images': Directory not empty
> ```

Makefile has been updated to avoid this issue (usually happens when a download is interrupted).

passalis:
Thanks for the useful comments and suggestions! I think I resolved all of them.

> But I would not put the C API in the projects folder but rather create a dedicated folder in the `src` for sources and in `include` (https://github.com/tasostefas/opendr_internal/tree/master/include) for headers and the tests should be part of the `tests` folder run by the CI.
> However, it is fine to keep the `projects/c_api/samples` folder as is.

I have moved source code of the C API into `src/c_api`, the header files into `include` and tests into `tests/sources/c_api`. I kept `projects/c_api/samples` folder as is. I also updated the tests into order to run automatically using GitHub actions using _check_. 

> We should also reduce the logs printed by the Makefile when installing the dependencies and cloning.

I reduced the verbosity of installation, cloning and compilation of third-party utilities. 

> By the way, everytime I type `make libopendr` the dependencies are reinstalled and the repositories and resources are downloade. This should be avoided and actions should only be executed when required and when something changed.

Makefile has been updated to avoid recompiling a target when it already exists. Also, installation has been moved into `install_compilation_dependencies`, while separate makefiles are used for the main dynamic library, tests and projects.

> I will also add style checks for C/C++ that are currently missing.

I checked using cppcheck and I updated the code in order to pass the tests. However, I think it would be better to first merge #113 and then this one to make sure that tests pass on CI.

passalis:
This PR is ready for re-review. However, cppcheck reports "cppcheck: error: could not find or open any of the paths given.". @stefaniapedrazzi should we ignore this?

stefaniapedrazzi:
> This PR is ready for re-review. However, cppcheck reports "cppcheck: error: could not find or open any of the paths given.". @stefaniapedrazzi should we ignore this?

I will check and fix it.

passalis:
Yes, it is ok!

dependencies/install_onnx.sh
```diff
@@ -0,0 +1,21 @@
+if [ ! -f /usr/local/lib/libonnxruntime.so ]; then
+
+  echo "Downloading onnxruntime (this might take a while) ..."
+  git clone -q --recursive https://github.com/Microsoft/onnxruntime
+  cd onnxruntime
+
+  # Checkout v1.6.0 (https://github.com/microsoft/onnxruntime/releases/tag/v1.6.0)
+  git checkout 718ca7f > /dev/null 2>&1
```
stefaniapedrazzi, line 8: Just a side note, this should be the same as doing directly 
```suggestion
  git checkout v1.6.0 > /dev/null 2>&1
```

stefaniapedrazzi (commented):
I'm having issues compiling onnxruntime on my local machine.
But do we really need to compile it from sources, or would be possible to download the compiled package from the onnxruntime repository?

stefaniapedrazzi:
I just fixed the cppcheck test and updated the C folder path.
Now that it is testing the correct directories there are some errors to fix.

.github/workflows/tests_sources.yml
```diff
@@ -82,4 +83,10 @@ jobs:
         make install_compilation_dependencies
         make install_runtime_dependencies
         pip install -r tests/sources/requirements.txt
-        python -m unittest discover -s tests/sources/tools/${{ matrix.package }}
+
+        if [ ${{ matrix.package }} = "ctests" ]; then
+          make libopendr
+          make ctests
```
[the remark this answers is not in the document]
passalis: Yes, I updated it.

passalis:
> I'm having issues compiling onnxruntime on my local machine.
> But do we really need to compile it from sources, or would be possible to download the compiled package from the onnxruntime repository?

I end up compiling ONNX, since I wasn't able to find any pre-compiled libraries for ONNX (for C/C++). They provide pre-compiled python packages, but I am not sure that they include all the functionality that we need to interface ONNX from other languages. Are there such libraries available on their repo? If there are such packages ready, I will update the PR and I fully agree that we should use them, since compilation takes some time and it is quite fragile process.

Regarding the errors, are using Ubuntu 20.04? I opted for v.1.6.0 ONNX since this was the latest version that was compiling with the already available build tools on 20.04 (the most recent ones required further changes). This version will most probably not work on earlier versions of Ubuntu.

stefaniapedrazzi:
Here it seems that there are precompiled packages.
But I might be wrong and I didn't checked in detail if all the required packages for the C API are included.
https://github.com/microsoft/onnxruntime/releases/tag/v1.6.0
GPU package: https://github.com/microsoft/onnxruntime/releases/download/v1.6.0/onnxruntime-linux-x64-gpu-1.6.0.tgz


> Regarding the errors, are using Ubuntu 20.04? I opted for v.1.6.0 ONNX since this was the latest version that was compiling with the already available build tools on 20.04 (the most recent ones required further changes). This version will most probably not work on earlier versions of Ubuntu.

Yes, I'm using Ubuntu 20.04.
I will try on a different Ubuntu 20.04 machine just to check.
However given that it works fine on the CI machine it seems fine.

passalis:
> Here it seems that there are precompiled packages.
> But I might be wrong and I didn't checked in detail if all the required packages for the C API are included.
> https://github.com/microsoft/onnxruntime/releases/tag/v1.6.0
> GPU package: https://github.com/microsoft/onnxruntime/releases/download/v1.6.0/onnxruntime-linux-x64-gpu-1.6.0.tgz

I am not sure how I managed to miss these... :flushed:  :smile:  I will check, but they seem just fine. I will update the PR to use these.

passalis:
I have updated the script to install the pre-compiled versions. Now it is much much faster and it seems to work correctly!

stefaniapedrazzi (approved):
Thank you!

Everything is fine for me.
It is also working fine on my local system now.

@ad-daniel: can you review of this PR?
Otherwise if any other partner wants to do the second review of this PR it is also fine.

ad-daniel (approved):
Looks good to me, thank you

PR #98, opened 2021-06-28
Data and Target setter/getter fixes

passalis:
This PR adds getters/setters for the basic fields of `Target` class, as well as specialized setters for the `Category` and `Pose` classes. It also contains a bugfix for the abstract `Data` class, where the setter was incorrectly implemented.


passalis:
This PR is ready for review (given that is passes all the tests :crossed_fingers: ). Every `Target`-derived class that does not implement custom type checking will use the setter/getters from the base `Target` class, so this PR should not affect the functionality of those methods. However, I think it would be good for everyone to re-check their `Target` and check if additional type checking needs to be implemented.

src/opendr/engine/target.py
```diff
@@ -68,6 +122,17 @@ def __init__(self, prediction: int, description=None, confidence=None):
         self.confidence = confidence
         self.description = description
 
+    @Target.data.setter
```
vniclas, line 125: Referencing the base class (`Target`) is inconsistent with how properties are being used in `data.py`.

src/opendr/engine/target.py
```diff
@@ -68,6 +122,17 @@ def __init__(self, prediction: int, description=None, confidence=None):
...
+    def data(self, data):
+        """
+        Setter for data. Category expects data of int type.
+        :param: data to be used for creating a Category object
+        """
+        if isinstance(data, int) or not data:
```
vniclas, line 131: This will actually allow other types than just integers. Did you mean to use `isinstance(data, int) or data is None`?

vniclas (changes requested):
Thanks for the proposed changes. I think there are still quite some inconsistencies in the code (also outside the lines of this PR) that should be addressed. All of them are easy fixes though.

src/opendr/engine/target.py
```diff
@@ -45,6 +45,60 @@ def __init__(self):
         self.confidence = None
         self.action = None
```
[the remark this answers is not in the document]
passalis: You are correct! Missed this!

src/opendr/engine/target.py
```diff
@@ -45,6 +45,60 @@ def __init__(self):
...
+        """
+        self._confidence = confidence
+
+    @property
+    def action(self):
+        """
+        Getter of action field.
+        This returns the selected/expected action.
+        :return: the action data held by the object
+        :rtype: Type of data
```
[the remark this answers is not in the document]
passalis: Fixed. Please note that the Action class does not yet exists (even though it is specified). It will be added in a future version of the toolkit (when active perception algorithms will be provided).

src/opendr/engine/target.py
```diff
@@ -68,6 +122,17 @@ def __init__(self, prediction: int, description=None, confidence=None):
         self.confidence = confidence
         self.description = description
```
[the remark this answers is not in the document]
passalis: I wanted to perform minimal changes to methods that I have not implemented. However, it seems to work after the changes, so I also fixed this.

src/opendr/engine/target.py
passalis: Fixed.

src/opendr/engine/target.py
passalis: Actually, this shouldn't be there in the first place (assignments with None do not make sense), so I completely removed it.

passalis:
> Thanks for the proposed changes. I think there are still quite some inconsistencies in the code (also outside the lines of this PR) that should be addressed. All of them are easy fixes though.

Hi @vniclas, thanks for noticing these! I have fixed them, along with a few more issues that I found.

src/opendr/engine/target.py
```diff
@@ -64,9 +119,61 @@ def __init__(self, prediction: int, description=None, confidence=None):
...
+
+    @property
+    def description(self):
+        """
+        Getter of description field.
+        This returns the description of the corresponding class.
+        :return: the description of the corresponding class
+        :rtype: str
+        """
+        return self.description
```
vniclas, line 141:
```suggestion
        return self._description
```

vniclas (changes requested):
I found one minor typo. Everything else looks good now. Thanks!

passalis:
> I found one minor typo. Everything else looks good now. Thanks!

Thanks again! I committed the change.

vniclas (dismissed):

docs/reference/engine-target.md
```diff
@@ -23,6 +23,11 @@ targets, this class serves as the basis for the more specialized forms of target
 All the classes should implement the corresponding setter/getter functions to ensure that the necessary
 type checking is performed (if there is no other technical obstacle to this, e.g., negative performance impact).
 
+`Target` provides the following fields that can be used by classes that inhert `Target`:
+- `data`, which holds the actual predictions/annotations,
+- `confidence`, which holds the prediction/annotation confidence,
+- `action`, which provides the expected active perception output.
```
stefaniapedrazzi, line 29: Just for consistency, fields of other Target classes are listed in italic when describing them:
```suggestion
- *data*, which holds the actual predictions/annotations,
- *confidence*, which holds the prediction/annotation confidence,
- *action*, which provides the expected active perception output.
```

src/opendr/engine/target.py
```diff
@@ -64,9 +119,61 @@ def __init__(self, prediction: int, description=None, confidence=None):
...
+        self._description = None
+
+        if prediction is not None:
+            self.data = prediction
+
+        if description is not None:
+            self.description = description
+
+        if confidence is not None:
+            self.confidence = confidence
```
stefaniapedrazzi, line 131: Is this correct or should it be
```suggestion
        if prediction is not None:
            self._data = prediction

        if description is not None:
            self._description = description

        if confidence is not None:
            self._confidence = confidence
```

stefaniapedrazzi (commented):
I think that if we go in this direction we should fix all the classes derived from `Target` otherwise this will be error prone and will produce unexpected results when using setters and getters.

* It seems that the `Category.__str__` method also needs not be updated.
Even with minimal changes `self.data` and `self.description` should now be changed to `self._data` and `self._description`.
However I would also change `self.category` to `self._category` for consistency.

* Some functions of derived classes (for example `Pose`) are still using `self.data` even if the the getters and setters are using `self._data` . It seems that this would cause issues.

Unrelated: I also spotted typos in `TrackingAnnotation` and `TrackingAnnotationList` with the spelling of `boudning_box` and `boudning_box_list`. I can fix them (as soon as possible) in a separate PR.

src/opendr/engine/target.py
passalis: This is intentional in order to perform type checking through the setter.

passalis:
I checked once again and I think we do have to change anything to the rest of the code - at least from a functional point of view - , since there is no external direct reference to protected fields like `._data` (this field didn't exist before this PR for `Target`-derived classes). All existing direct references (e.g., `object.data`) will now refer to either the base setter and getter implemented in the `Target` or to the overridden setter and getter provided in the concrete implementation (if available). This allows to perform type checking and any other additional check in an easy and consistent way for any other class in the hierarchy.

I agree that we should try to perform type checking in all `Target` and `Data`-related classes to catch errors early and ensure that all algorithms of the same kind receive the same type of data. If we do not perform type checking there, the errors (if we use an unsupported type of data) will most probably arise inside the actual implementation of an algorithm - and in some case the error message might not be so helpful to understand what went wrong (pytorch and tensorflow are not always helpful with their error messages :yum: ).

stefaniapedrazzi (approved):
Thank you!

Some tests seems are failing (even if not reported by the CI system).
However I get the same on the master branch so they should be unrelated to this PR.

passalis:
The tests are now failing, but I get the same behavior also on master. The failure also does not seem to be related to changes made in this PR.

passalis:
> The tests are now failing, but I get the same behavior also on master. The failure also does not seem to be related to changes made in this PR.

I see that now the tests pass, without any change and despite still failing locally. Perhaps `rtol` and `atol` should be increased. I will open an issue regarding this, since it is not related to this PR, it happens on the master branch also and it occurred in other PRs as well.

vniclas:
Will this PR be merged soon so that it can be considered in the changes of PR #96?

passalis:
Hi, I merged master into this branch and I am running the tests again, so as soon the tests pass and we have 1 more approval I will merge this.

vniclas (approved):

PR #87, opened 2021-06-08
Speech command recognition - Quadratic Self-ONN

jovaroo:
Adds an implementation of a Quadratic Self-ONN based learner, which can be used for keyword spotting or other limited-vocabulary speech recognition tasks.

PR #104, opened 2021-07-11
Pose estimation NumPy compatibility fixes

passalis:
This PR updates the requirements of pose estimation in order to use a patched version of the (already modified in order to work on windows) pycocotools. This fixes the issue reported in #87 and makes the implementation compatible with any NumPy version.


passalis:
This is ready for review.

stefaniapedrazzi (approved):
@jovaroo could you please also review this PR and give a feedback if this is fixing your issue with numpy compatibility?

jovaroo (approved):
The `int` wrappers should indeed help with numpy compatibility, and in no case should this cause any harm.

PR #113, opened 2021-07-21
Add C/C++ test using cppcheck

stefaniapedrazzi:
Run cppcheck on C/C++ files and update the version if clang-format.

### DIFF
--- a/tests/test_cppcheck.py
+++ b/tests/test_cppcheck.py
@@ -1,0 +1,138 @@
+#!/usr/bin/env python
+
+# Copyright 2020-2021 Cyberbotics Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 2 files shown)


stefaniapedrazzi:
Tests passes after disabling the `object_tracking_2d` tests that are currently failing on master.

stefaniapedrazzi:
I would temporary disable the `object_tracking_2d` as it is breaking the tests on all the PRs.

tests/test_cppcheck.py
```diff
@@ -0,0 +1,138 @@
...
+                    for sourceDir in sourceDirs:
+                        if line.startswith(sourceDir):
+                            shouldSkip = False
+                            for skipped in skippedDirs + skippedfiles:
+                                if line.startswith(skipped):
+                                    shouldSkip = True
+                                    break
+                            if not shouldSkip:
+                                command += ' \"' + line + '\"'
+                            continue
```
vniclas, line 77: This `continue` does not have any effect.

vniclas (approved):

vniclas (changes requested):
There are errors reported by `cppcheck` but the tests are not failing.

stefaniapedrazzi:
> There are errors reported by `cppcheck` but the tests are not failing.

For the moment we don't have any C/C++ files to check and this is what `cppcheck` is reporting .
I checked and errors in C/C++ makes the test failing correctly.
Moreover as soon as at least one C/C++ file is found the current errors disappear.
So it is correct that the current errors, even if printed in the log, don't make the test fail.

vniclas:
Thanks for following up on this!

vniclas (approved):

vniclas (approved):

passalis (approved):
Thank you! Seems fine.

PR #120, opened 2021-08-30
Fix gpu support in Video-based Human Activity Recognition

LukasHedegaard:
This PR fixes the bug in `x3d_learner` and `cox3d_learner`, that `learner.model` was not assigned to the correct device. With this PR, cuda devices will be supported by the these learners.

The Activity Recognition demo has also been updated to support gpu.


passalis (approved):
This seems fine to me. Thank you!

passalis:
> This seems fine to me. Thank you!

I just noticed the cancelled test run. I think you could try re-running the tests to see if this persists, since github is not always consistent with testing time... I do not see any change that could affect test time on the current setup.

LukasHedegaard:
It seems that we need #125 to be integrated before the tests for this PR will pass.

LukasHedegaard:
Despite changing the test workflow #125 , the runs in this PR are still cancelled after 6 hours. I will close this PR and re-open a new one to see if it fixes the issue.

Issue #123, opened 2021-09-03
Tests cancel after 360 m


LukasHedegaard:
With a growing toolkit size, CI tests are naturally taking longer and longer. However, the test duration has now reached a point, where GitHub Actions cancels tests before they finish (after 360 minutes). This makes integration of bug-fixes or new tools impossible in the current setup, even through the tests themselves do not have errors.

A "hack" to circumvent this could be to split the Test API workflow into multiple sub-workflows, each of which test a subset of the methods, e.g. by replacing the last line in "opendr_internal/.github/workflows/tests_sources.yml" ( `python -m unittest discover -s tests/sources`) with one or more targeted tests (e.g. `python -m unittest discover -s tests/sources/tools/perception/activity_recognition`.)

stefaniapedrazzi:
You are welcome to go ahead and implement this solution.
We could add a script in the test folder that automatically loops through all the available folders and launches the tests.

Issue #124, opened 2021-09-06
Error on tests on object_tracking_3d_ab3dmot


charsyme:
ERROR: test_eval (tools.perception.object_tracking_3d.ab3dmot.test_object_tracking_3d_ab3dmot.TestObjectTracking3DAb3dmot)

  File "/home/runner/work/opendr_internal/opendr_internal/src/opendr/perception/object_tracking_3d/ab3dmot/algorithm/evaluate.py", line 475, in compute
    t = seq_tracker[f]
IndexError: list index out of range

charsyme:
@passalis I think it was fixed that period

passalis:
Ok, so I think we can close this.

PR #125, opened 2021-09-07
Split API test into sub-tests

LukasHedegaard:
This PR implements the solution discussion in #123 by splitting the API tests into multiple independent tests.


LukasHedegaard:
As seen in the build results above, the total test time is considerably lower now due to test parallelisation.

A downside to the workflow modification is, that all new tools must add their path to `.github/workflows/tests_sources.yml`.
This should be communicated clearly to contributors and added to the [PR checklist](https://github.com/tasostefas/opendr_internal/wiki/Adding-New-Tool#pr-checklist) as well.

vniclas:
This is golden as it also provides a solution to the current issue in PR #122

LukasHedegaard:
@stefaniapedrazzi @passalis, the PR is ready for review.

Once again, I would like to underline that all new tools must add their path to `.github/workflows/tests_sources.yml`.
This should be communicated clearly to contributors and added to the [PR checklist](https://github.com/tasostefas/opendr_internal/wiki/Adding-New-Tool#pr-checklist) as well.

.github/workflows/tests_sources.yml
```diff
@@ -45,14 +45,41 @@ jobs:
...
+    - name: Set up Python 3.8
+      if: github.event.pull_request.draft == false
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Test Tools
+      if: ${{ contains(github.event.pull_request.labels.*.name, 'test tools') }}
       run: |
-        ${{ matrix.DEPENDENCIES_INSTALLATION }}
+        sudo apt -y install clang-format-10 cppcheck
```
stefaniapedrazzi, line 78: Why do you need clang format and cppcheck for testing the functionality?
These tools should only be run in the `test-sources` job to check the code sources.

.github/workflows/tests_sources.yml
```diff
@@ -45,14 +45,41 @@ jobs:
...
+          - perception/compressive_learning
+          - perception/face_recognition
+          - perception/heart_anomaly_detection
+          - perception/multimodal_human_centric
+          - perception/object_tracking_2d
+          - perception/object_detection_3d
+          - perception/pose_estimation
+          - perception/speech_recognition
+          # - perception/object_tracking_3d
+    runs-on: ubuntu-20.04
```
stefaniapedrazzi, line 66: Note that at some point we will run the tests on others OS too.
But this can be changed later.

.github/workflows/tests_sources.yml
```diff
@@ -45,14 +45,41 @@ jobs:
...
+          - perception/activity_recognition
+          - perception/compressive_learning
+          - perception/face_recognition
+          - perception/heart_anomaly_detection
+          - perception/multimodal_human_centric
+          - perception/object_tracking_2d
+          - perception/object_detection_3d
+          - perception/pose_estimation
+          - perception/speech_recognition
+          # - perception/object_tracking_3d
```
stefaniapedrazzi, line 65: Why is this commented out?

.github/workflows/tests_sources.yml
```diff
@@ -45,14 +45,41 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+      if: github.event.pull_request.draft == false
```
stefaniapedrazzi, line 69: Not sure it make sense any longer to skip this in case of draft PR.
Does it work if you set the `test tools` label for a draft PR?

.github/workflows/tests_sources.yml
```diff
@@ -45,14 +45,41 @@ jobs:
...
+    - name: Set up Python 3.8
+      if: github.event.pull_request.draft == false
```
stefaniapedrazzi, line 71: Not sure it make sense any longer to skip this in case of draft PR.
Does it work if you set the `test tools` label for a draft PR?

stefaniapedrazzi (commented):
Thank you.

I have a couple of comments.
Once merged we can update the instructions on the wiki.

.github/workflows/tests_sources.yml
vniclas: I would also vote for enabling tests for draft PR as well. This way, one can check whether tests are passing before removing the _draft_ status.

.github/workflows/tests_sources.yml
LukasHedegaard: There is an open issue for the object tracking tests #124. 
To fast-track this PR and thereby other PRs, which wait for this fix, I excluded it. I assume #124 will be handled in a separate bug-fix.

.github/workflows/tests_sources.yml
LukasHedegaard: Yes, it is trivial to extend the matrix runner in the samme manner as it was done for the 'test-sources' workflow

.github/workflows/tests_sources.yml
LukasHedegaard: Removed in commit 77abe5f41c

.github/workflows/tests_sources.yml
LukasHedegaard: Removed in commit 77abe5f41c

.github/workflows/tests_sources.yml
LukasHedegaard: You are right, it probably doesn't. I just replaced the `${{ matrix.DEPENDENCIES_INSTALLATION }}` step with the contents for ubuntu 20.04. This was done to keep the test steps the same as in the old version. 
The line was remove in commit dbd212689ece9ffd

.github/workflows/tests_sources.yml
LukasHedegaard: I have update the code to use a os matrix (with ubuntu 20.04 only) in commit 89b95cb2

stefaniapedrazzi (approved):
Thank you

passalis (approved):
Thank you!

PR #126, opened 2021-09-13
Fix gpu support in Video-based Human Activity Recognition

LukasHedegaard:
This PR fixes the bug in `x3d_learner` and `cox3d_learner`, that `learner.model` was not assigned to the correct device. With this PR, cuda devices will be supported by the these learners.

The Activity Recognition demo has also been updated to support gpu.


passalis (commented):
This PR seems fine. Do we also need to update the documentation to reflect this change?

LukasHedegaard:
> This PR seems fine. Do we also need to update the documentation to reflect this change?

The PR only contains a bug-fix, so the docs should not have changed.

stefaniapedrazzi (approved):

passalis (approved):

passalis (commented):
Thank you for the quick fix!

PR #140, opened 2021-10-15
Fix broken CI due to pyYAML upgrade

ad-daniel:
Fixes https://github.com/tasostefas/opendr_internal/issues/139

In pyYAML, the load method requires to have a Loader being passed. Loading without this argument was already deprecated before, but with version 6 it is enforced. The change broke `rosinstall` module and considering that according to their [git](https://github.com/vcstools/rosinstall) the repository has been archived due to a lack of maintainance, I don't expect them to fix it anytime soon. As such, it was replaced by vcstool.

### DIFF
--- a/.github/workflows/tests_sources.yml
+++ b/.github/workflows/tests_sources.yml
@@ -79,10 +79,10 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


vniclas (approved):
Looks good to me (assuming that all tests will pass). Thanks for addressing the issue.

PR #151, opened 2021-11-12
Object detection 3d demo

iliiliiliili:
3D object detection demo. Works with disk data, O3M Lidar and RPLidar.


ad-daniel (commented):
Thank you!

Is any of these datasets freely available? I mean, could they be included directly or downloaded from the FTP? Having them behind a website that requires registering an account doesn't sound to me very user friendly, especially since it's meant to be a demo. I'm not from your field, so I don't know how common access to these datasets actually is, so please correct me if I'm wrong.

Concerning the `*.proto`, are all of them needed? I would keep only those that are being referenced by the demo (or the README should mention they exist, otherwise I don't see much point in having them.

iliiliiliili:
Thank you @ad-daniel. The dataset that is used in demo is a KITTI dataset and is the most used dataset in this field (it is also used for 3d detection and tracking methods in OpenDR). It was possible to download it without registration earlier, but now you have to register. We have two small subsets of it (nano_kitti and mini_kitti) in FTP, but I don't think we should (and can) provide the whole dataset there.

Protos included are those that I used for the OpenDR presentation, so it is expected that it would be useful for others to use any of them.

passalis:
Thank you! Seems ok to me too. 

I have only one concern regarding the re-distribution of fonts (I see there is an `arial.ttf` inside the files). I do not know the terms of redistribution of the specific font (usually embedding is allowed for documents, but I think standalone re-distribution in apps in not allowed for windows fonts). Can you check if there is a specific term for Arial?

iliiliiliili:
Thank you, @passalis, I removed the ttf file and made it work without it for this demo.

ad-daniel (approved):

passalis (approved):
Thank you!

PR #152, opened 2021-11-24
Semantic segmentation ros

mtzelepi:
Semantic Segmentation Ros Node


passalis (approved):
Thank you for your contribution! I think this is fine.

ad-daniel (approved):
Thank you, when possible it would be good to have an automatic test for ROS interface as well.

PR #155, opened 2021-11-29
ROS node for heart anomaly detection

katerynaCh:
This PR adds a ROS node for heart anomaly detection for both Neural bag-of-feature and Gated Recurrent Unit models.

### DIFF
--- a/projects/opendr_ws/src/perception/scripts/heart_anomaly_detection.py
+++ b/projects/opendr_ws/src/perception/scripts/heart_anomaly_detection.py
@@ -1,0 +1,106 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# Copyright 2020-2021 OpenDR European Project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


projects/opendr_ws/src/perception/scripts/heart_anomaly_detection.py
```diff
@@ -0,0 +1,105 @@
...
+        rospy.loginfo("Heart anomaly detection node started!")
+        rospy.spin()
+
+    def callback(self, msg_data):
+        """
+        Callback that process the input data and publishes to the corresponding topics
+        :param data: input message
+        :type data: std_msgs.msg.Float32MultiArray
+        """
+        # Convert Float32MultiArray to OpenDR Timeseries
```
passalis, line 81: Would it be good to provide this function as part of the bridge? It's just a two lines and it probably also needs to receive the corresponding arguments as input. However, I think it would provide greater consistency with the rest of the nodes.

passalis (commented):
Thank you! This seems fine. I have only one minor concern regarding the translation between Float32MultiArray and OpenDR data types, which could be addressed.

passalis (approved):
Thanks!

ad-daniel (approved):
Thank you!

passalis (approved):

ad-daniel (approved):

PR #158, opened 2021-12-02
Skeleton based action recognition ros

negarhdr:
This pull request adds the ROS node of skeleton-based human action recognition (both baseline and progressive versions). 
The script is added in 
`projects/opendr_ws/src/perception/scripts/skeleton_based_action_recognition.py`

and a short description is added in: 
`projects/opendr_ws/src/perception/README.md`

and the list of nodes is updated in:
`projects/opendr_ws/README.md`

PR #133, opened 2021-10-12
Simplify import statements 

passalis:
This PR aims to simplify the import statments by exposing the commonly used classes and functions in the each module's `__init.py__` file following #132.

My suggestion is to go from:

```python
from opendr.perception.pose_estimation.lightweight_open_pose.lightweight_open_pose_learner import \
    LightweightOpenPoseLearner
```

to simply

```python
from opendr.perception.pose_estimation import LightweightOpenPoseLearner
```

PR #159, opened 2021-12-02
Landmark based facial expression recognition ros

negarhdr:
This pull request adds the ROS node of landmark-based facial expression recognition.
The script is added in
projects/opendr_ws/src/perception/scripts/landmark_based_facial_expression_recognition.py

and a short description is added in:
projects/opendr_ws/src/perception/README.md

and the list of nodes is updated in:
projects/opendr_ws/README.md


passalis (commented):
Thank you! The ROS functionality seems fine. However, I have similar concerns with #158, since the same logic is also followed here. Furthermore, you could also consider whether `_landmark2numpy` and `_data_normalization` should be part of the main toolkit, e.g., similar to the way ``gen_muscle_data`` are.

projects/opendr_ws/src/perception/scripts/landmark_based_facial_expression_recognition.py
```diff
@@ -0,0 +1,174 @@
...
+import pandas as pd
+from std_msgs.msg import String
+from vision_msgs.msg import ObjectHypothesis
+from sensor_msgs.msg import Image as ROS_Image
+from opendr_bridge import ROSBridge
+from opendr.perception.facial_expression_recognition.landmark_based_facial_expression_recognition import \
+     ProgressiveSpatioTemporalBLNLearner
+from opendr.perception.facial_expression_recognition.landmark_based_facial_expression_recognition.\
+     algorithm.datasets.landmark_extractor import landmark_extractor
+from opendr.perception.facial_expression_recognition.landmark_based_facial_expression_recognition.\
```
passalis, line 29: Please simplify the import statements as described in #133. If needed you can update the corresponding `__init__.py` files

negarhdr: The imports are simplified now.

negarhdr:
> Thank you! The ROS functionality seems fine. However, I have similar concerns with #158, since the same logic is also followed here. Furthermore, you could also consider whether `_landmark2numpy` and `_data_normalization` should be part of the main toolkit, e.g., similar to the way `gen_muscle_data` are.

data normalization is part of the main toolkit so I removed it from the ROS node implementation. but landmark2numpy is only for demo and ROS node where we get a single image as input and want to process its landmarks. So I think it should be kept as it is here.

negarhdr:
@ad-daniel I see that the tool test (ubuntu-20.04, engine) is failing for this PR. I am not sure if there is something wrong from my side or not.

ad-daniel:
@negarhdr I don't see anything wrong with the PR itself. I'll try to rerun it. I'm also running another test to make sure nothing is wrong with the test suite.

ad-daniel:
The other issue appears to be fixed, what fails now seems related to this PR

negarhdr:
> The other issue appears to be fixed, what fails now seems related to this PR

Yes. I need to add two more dependencies (imutils and dlib) and it seems the error is related to installing these packages.

passalis (approved):
Thank you for the changes! I think this PR is ready as soon as the tests issue is fixed.

ad-daniel (approved):
Thank you.

passalis (approved):

PR #164, opened 2021-12-07
Image input consistency for OpenDR Tools

passalis:
This PR updates the toolkit to ensure that all tools that uses images receive images formatted in the same way.

The following tasks should be implemented to complete this PR:
- [x] Update Image to support NCHW/RGB
- [ ] Update implementations to support the updated Image
- [ ] Update rosbridge
- [ ] Update ROS nodes

Tools that might need to be updated (please check if you updated and/or already compliant with the changes):

- [ ] Face Detection
- [ ] Face Recognition
- [ ] Landmark based facial expression recognition
- [ ] Human Pose Estimation
- [ ] Online Video Activity Recognition
- [ ] Semantic Scene segmentation 
- [ ] Object detection 2D
- [ ] Efficient Panoptic Segmentation
- [ ] Object Tracking 2D
- [ ] Multimodal Human Centric Perception
- [ ] Multimodal 2D object detection
- [ ] Human Model Generation

ROS nodes that might need to be updated (please check if you updated and/or already compliant with the changes):
- [] Pose Estimation
- [] 2D Object Detection
- [] Face Detection
- [] Panoptic Segmentation
- [] Face Recognition
- [] Semantic Segmentation
- [] RGBD Hand Gesture Recognition


iliiliiliili:
Should we add functions to return data in channel first/last format? This way, it will be easy to just call `input: np.ndarray = image.channel_first()` and not care about the internal order.

passalis:
I added `convert()` function which allows for easily changing between different formats. This is supposed to be mainly used internally (e.g., you check how it is used by pose estimation) to bring the data into the appropriate format easily. For now, I also kept `opencv()`, which overlaps to some extent which this function (I will clean the code later today by re-using the other implementation), in order to provide an easy to use interface for visualization (which will be probably more useful for end users than the `convert()` function).

src/opendr/engine/data.py
```diff
@@ -211,15 +211,32 @@ def __str__(self):
 class Image(Data):
     """
     A class used for representing image data.
+    OpenDR uses NCHW/RGB conventions
     This class provides abstract methods for:
     - returning a NumPy compatible representation of data (numpy())
+    - loading an input directly into OpenDR compliant format (open())
+    - getting an image into OpenCV- compliant format (opencv()) for visualization purposes
```
vniclas, line 218: **ToDo:** There's a space between _OpenCV-_ and _compliant_ that has to be removed.

src/opendr/engine/data.py
```diff
@@ -211,15 +211,32 @@ def __str__(self):
...
     """
 
-    def __init__(self, data=None, dtype=np.uint8):
+    def __init__(self, data=None, dtype=np.uint8, guess_format=True):
+        """
+        Image constructor
+        :param: dtype type of the image data provided
+        :guess_format: try to automatically guess the type of input data and convert them to OpenDR format
```
vniclas, line 225: This doesn't adhere to the usual way of documentation. It should be something along these lines:

```
:param data: blablabla
:type data: numpy-array-like
:param dtype: ....
:type dtype: data-type
:param guess_format: ...
:type guess_format: bool
```

src/opendr/engine/data.py
```diff
@@ -211,15 +211,32 @@ def __str__(self):
 class Image(Data):
     """
     A class used for representing image data.
+    OpenDR uses NCHW/RGB conventions
```
vniclas, line 214: Out of curiosity, what about grayscale images?

src/opendr/engine/data.py
```diff
@@ -283,8 +300,42 @@ def open(cls, filename):
             raise FileNotFoundError('The image file does not exist.')
         data = cv2.imread(filename)
         data = cv2.cvtColor(data, cv2.COLOR_BGR2RGB)
+        data = np.transpose(data, (2, 0, 1))
```
vniclas, line 303: Maybe add a comment stating what's being done here.

src/opendr/engine/data.py
```diff
@@ -283,8 +300,42 @@ def open(cls, filename):
...
+
+    def convert(self, format='channels_first', channel_order='rgb'):
+        """
+        Returns the data in channels first/last format using either 'rgb' or 'bgr' ordering
+        :param format: either 'channels_first' or 'channels_last'
+        :param channel_order: either 'rgb' or 'bgr'
+        :return an image with the appropriate format
+        :rtype NumPy array
+        """
+        data = self.data
```
vniclas, line 326: This has to be a copy instead of a reference!

src/opendr/engine/data.py
```diff
@@ -283,8 +300,42 @@ def open(cls, filename):
...
+        :rtype: NumPy array
+        """
+        data = np.transpose(self.data, (1, 2, 0))
+        data = cv2.cvtColor(data, cv2.COLOR_RGB2BGR)
+        return data
+
+    def convert(self, format='channels_first', channel_order='rgb'):
+        """
+        Returns the data in channels first/last format using either 'rgb' or 'bgr' ordering
+        :param format: either 'channels_first' or 'channels_last'
```
vniclas, line 321: Both params miss the `type ...` specification.

vniclas (changes requested):
I started commenting with single comments but then realized that there is more to be changed.

projects/opendr_ws/src/ros_bridge/src/opendr_bridge/bridge.py
```diff
@@ -49,24 +48,21 @@ def from_ros_image(self, message: ImageMsg, encoding: str='passthrough') -> Imag
...
         :type image: engine.data.Image
         :param encoding: encoding to be used for the conversion (inherited from CvBridge)
         :type encoding: str
         :return: ROS image
         :rtype: sensor_msgs.msg.Image
         """
-        # Convert from the OpenDR standard (RGB) to OpenCV standard (BGR)
-        bgr_image = cv2.cvtColor(image.numpy(), cv2.COLOR_RGB2BGR)
-        message = self._cv_bridge.cv2_to_imgmsg(bgr_image, encoding=encoding)
+        # Convert from the OpenDR standard (NCHW/RGB) to OpenCV standard (NHWC/BGR)
```
vniclas, line 64: There are no batches (N) involved here, right?

src/opendr/engine/data.py
passalis: Thanks for noticing. I was going to clean up the code later today or tomorrow.

src/opendr/engine/data.py
passalis: Given that all implementations currently work with color images, I do not see any immediate need to support grayscale images. I can be done though quite easily by checking if image has only one channel

src/opendr/engine/data.py
passalis: Thanks, I will add one.

src/opendr/engine/data.py
passalis: I would like to avoid as much as possible copy operations on this function that is envisioned to be used by OpenDR implementations to easily change between different formats to avoid double conversions just to follow OpenDR format. Users are supposed to use `numpy()` to get a reference to internal data and `opencv()` to get a copy for visualization. So, indeed, this function may or may not copy the memory depending on the situation. Perhaps we should explicitly state this to avoid cases where users are not aware of this or even hide this function from the external Image interface.

projects/opendr_ws/src/ros_bridge/src/opendr_bridge/bridge.py
passalis: This is correct! I will update this.

src/opendr/engine/data.py
vniclas: I realize now that it is okay since both `np.transpose()` and `cv2.cvtColor()` return copies instead of references. Thus, we do not have to create an additional copy in this line if we just want to avoid changing the internal `data` representation. However, it is not so nice to change the returning behavior based on the input arguments without being intended by the user.

Strictly speaking, `numpy()` should rather return a copy as well since the `data` property might not be a numpy array for the generic `Data` class. Hence, for consistency with those inherited types, which do not use an internal numpy class for their `data` property, the `numpy()` method of the `Image` type should probably rather return a copy as well.

passalis: I fully agree and - indeed - this is a way to introduce nasty bugs for the end users... So, for `numpy()` we should probably return a copy (I will update it), since it is tailored to the end users. My only concern for `convert()` is that we add an additional overhead even for cases where `convert()` wouldn't do any conversion. Since `convert()` is used by various Learners this will probably add an overhead (which - I do not know if it really matters). We could force `convert()` to always copy the data just to be on the safe side.

vniclas: Maybe, we should change the `numpy()` method in a separate PR as this might not only affect `Image` (most probably).

Some options that I can think of:
1) State it in the documentation.
2) Always return a copy.
3) Add a parameter `enforce_copy` to always return a copy or use the current implementation, i.e., return a reference if converting is not required.

vniclas: Sounds good.

katerynaCh:
The multimodal human centric update seems fine, thank you! The ROS node was updated accordingly. Compressive learning seems fine, too, but maybe @viebboy could confirm netter.

projects/opendr_ws/src/perception/scripts/rgbd_hand_gesture_recognition.py
```diff
@@ -103,8 +103,8 @@ def preprocess(self, image, depth_img):
         :param depth_data: input depth image
         :type depth_data: engine.data.Image
         '''
-        image = image.numpy() / (2**8 - 1)
-        depth_img = depth_img.numpy() / (2**16 - 1)
+        image = image.convert(format='channels_last') / (2**8 - 1)
```
passalis, line 106: Should we also add here channel_order="bgr" since we feed this image into cv2.resize in l.110?

katerynaCh: I am not sure what you mean? cv2.resize does not change the channel ordering, so to my understanding this is fine

passalis: You are correct! It is indeed fine.

passalis:
This PR is ready for review. I tried to avoid double conversions, but I didn't want to change the implementations too much and risk introducing bugs, so in the vast majority of the cases, I only added conversion statements in the corresponding learners.

src/opendr/perception/face_recognition/face_recognition_learner.py
```diff
@@ -455,7 +455,7 @@ def infer(self, img):
         """
         if not isinstance(img, Image):
             img = Image(img)
```
vniclas, line 457: This is a bit dangerous. I would rather raise a TypeError.

projects/perception/lightweight_open_pose/demos/inference_demo.py
```diff
@@ -44,12 +45,13 @@
...
     pose_estimator.download(path=".", mode="test_data")
     image_path = join("temp", "dataset", "image", "000000000785.jpg")
-    img = cv2.imread(image_path)
+    img = Image.open(image_path)
 
     if onnx:
         pose_estimator.optimize()
 
     poses = pose_estimator.infer(img)
+    img = img.opencv()
```
vniclas, line 54: This line overwrites a variable with a new type. Obviously, this is not a bug but it might be a bit confusing.

src/opendr/perception/compressive_learning/multilinear_compressive_learning/multilinear_compressive_learner.py
```diff
@@ -405,7 +405,7 @@ def infer_from_compressed_measurement(self, measurement):
         self.model.to(torch.device(self.device))
         self.model.eval()
 
-        measurement = measurement.numpy()
+        measurement = measurement.convert("channels_last")
```
vniclas, line 408: Is this missing an explicit conversion to RGB similar to the lines above? Alternatively, remove the parameter above as it is the default anyway.

src/opendr/perception/multimodal_human_centric/rgbd_hand_gesture_learner/algorithm/data.py
```diff
@@ -65,7 +65,7 @@ def __len__(self,):
     def __getitem__(self, i):
         x, y = self.dataset.__getitem__(i)
         # change from rows x cols x channels to channels x rows x cols
-        x = np.transpose(x.numpy(), axes=(2, 0, 1))
+        x = x.convert("channels_first")
```
vniclas, line 68: Are statements like this be left on purpose? They seem to be a bit useless.

src/opendr/perception/object_detection_2d/centernet/centernet_learner.py
```diff
@@ -395,16 +395,9 @@ def infer(self, img, threshold=0.2, keep_size=True):
...
-        elif isinstance(img, np.ndarray):
-            _img = img
-        else:
-            if img is not None:
-                error_msg = "Input should be of type Image or numpy array."
-            else:
-                error_msg = "Input image is of type None."
-            raise ValueError(error_msg)
+        if not isinstance(img, Image):
+            img = Image(img)
```
vniclas, line 399: This doesn't exactly reproduce the previous code.

src/opendr/perception/object_detection_2d/ssd/ssd_learner.py
```diff
@@ -555,12 +555,9 @@ def infer(self, img, threshold=0.2, keep_size=False):
...
         self._model.set_nms(nms_thresh=0.45, nms_topk=400)
 
-        if isinstance(img, Image):
-            _img = img.numpy()
-        elif isinstance(img, np.ndarray):
-            _img = img
-        else:
-            raise ValueError("Input should be of type Image or numpy array.")
+        if not isinstance(img, Image):
+            img = Image(img)
```
vniclas, line 559: This doesn't reproduce the previous code.

src/opendr/perception/object_detection_2d/yolov3/yolov3_learner.py
```diff
@@ -431,12 +431,9 @@ def infer(self, img, threshold=0.1, keep_size=True):
...
         self._model.set_nms(nms_thresh=0.45, nms_topk=400)
 
-        if isinstance(img, Image):
-            _img = img.numpy()
-        elif isinstance(img, np.ndarray):
-            _img = img
-        else:
-            raise ValueError("Input should be of type Image or numpy array.")
+        if not isinstance(img, Image):
+            img = Image(img)
```
vniclas, line 435: This doesn't reproduce the previous code.

src/opendr/perception/object_tracking_2d/deep_sort/object_tracking_2d_deep_sort_learner.py
```diff
@@ -418,7 +418,8 @@ def _prepare_datasets(
         def map_dataset(dataset):
 
             def image_to_pil(image: Image):
```
vniclas, line 420: Eventually, we should probably make this generic. E.g., I have a similar function in my code.

vniclas (changes requested):
Thanks @passalis.
I have some minor comments. Some of them might just require a written reply instead of code changes.

docs/reference/engine-data.md
```diff
@@ -91,9 +93,19 @@ The [Image](#class_engine.data.Image) class has the following public methods:
...
   Given that *data* argument is already internally stored in [NumPy](https://numpy.org)-compatible format, this method is equivalent to `data()`.
 
+#### opencv()
+  Return an [OpenCV](https://opencv.org)-compatible representation of data.
+  This method transforms the internal CHW/RGB representation into HWC/BGR used by OpenCV.
+
 #### open(filename)
   Construct a new [Image](#class-engine.data.Image) object from the given image file.
 
+#### convert(format, channel_order)
```
ad-daniel, line 103:
```suggestion
#### convert(format='channels_first', channel_order='rgb')
```

ad-daniel: we should probably specify in the docs the default parameters

projects/perception/face_recognition/demos/inference_demo.py
```diff
@@ -33,8 +33,8 @@
 
     recognizer.download(path=".", mode="test_data")
     recognizer.fit_reference(path=join(".", "test_data", "images"), save_path=".")
-    image_path = join(".", "test_data", "images", "Mr. Bean", "1.jpg")
-    img = cv2.imread(image_path)
```
ad-daniel, line 37: We should probably adapt the examples provided in the documentation accordingly too, but isn't critical

ad-daniel (changes requested):

src/opendr/perception/face_recognition/face_recognition_learner.py
passalis: Indeed. However, this is already extensively used in other implementations to allow for handling OpenCV arrays without adding boilerplate code. I will add an extra check to ensure that data can be cast into images directly into the constructor in order to raise the error from there.

src/opendr/perception/compressive_learning/multilinear_compressive_learning/multilinear_compressive_learner.py
passalis: This is a bit strange case. We no longer have RGB channels, since this is a compressed form of the image (indeed, in testing 2 channels are used). By using the default parameters we do not perform any RGB2BGR conversion, but we indeed get a different channel ordering.

projects/perception/lightweight_open_pose/demos/inference_demo.py
passalis: Fixed.

src/opendr/perception/multimodal_human_centric/rgbd_hand_gesture_learner/algorithm/data.py
passalis: Again, we do not specify channels order here, since we work with RGBD images.

src/opendr/perception/compressive_learning/multilinear_compressive_learning/multilinear_compressive_learner.py
vniclas: Got it. Thanks for the explanation.

src/opendr/perception/object_detection_2d/centernet/centernet_learner.py
passalis: Code added to ensure that valid types are passed into Image constructor.

projects/perception/face_recognition/demos/inference_demo.py
passalis: I think I updated any `cv2.imread()` instances. I also fixed some old `OpenDR` import statements (this PR is now a bit messy...).

vniclas:
Let us know once we should have another look at this.

ad-daniel:
Looks like one of the tests fails because of this
`"/perception/skeleton_based_action_recognition/pretrained_models/stgcn/stgcn_nturgbd_cv_joint-49-29400": directory not found.`

@negarhdr indeed it seems this file no longer is present in the FTP, was it recently removed/renamed?

src/opendr/perception/object_detection_2d/yolov3/yolov3_learner.py
passalis: Fixed by improving constructor.

src/opendr/perception/object_tracking_2d/deep_sort/object_tracking_2d_deep_sort_learner.py
passalis: Perhaps we could add an "pil()" method to Image and also try to perform a cleanup in the code to reduce duplicate conversions that I think happen in some learners.

passalis:
Thank you both for the detailed comments! I think it is completed from my side. Let me know if I missed something or If you find anything else that needs to be addressed. 

@ad-daniel We are looking into the issue with the missing files.

docs/reference/detr.md
```diff
@@ -328,16 +335,16 @@ Furthermore, demos on performing [training](../../projects/perception/object_det
   import numpy as np
   import urllib
   import cv2
-  from opendr.perception.object_detection_2d.detr.detr_learner import DetrLearner
-  from opendr.perception.object_detection_2d.detr.algorithm.util.draw import draw
-  
-  
+  from opendr.perception.object_detection_2d import DetrLearner
+  from opendr.perception.object_detection_2d import draw
```
jelledouwe, line 339:
```suggestion
  from opendr.perception.object_detection_2d.detr.algorithm.util.draw import draw
```
The draw utility function is not specified in the __init__.py in object_detection, so that would result in an error.
It can be imported like this or is it better to add it to the __init__.py?

docs/reference/detr.md
```diff
@@ -354,15 +361,15 @@ Furthermore, demos on performing [training](../../projects/perception/object_det
   import numpy as np
   import urllib
   import cv2
-  from opendr.perception.object_detection_2d.detr.detr_learner import DetrLearner
-  from opendr.perception.object_detection_2d.detr.algorithm.util.draw import draw
-  
+  from opendr.perception.object_detection_2d import DetrLearner
+  from opendr.perception.object_detection_2d import draw
```
jelledouwe, line 365:
```suggestion
  from opendr.perception.object_detection_2d.detr.algorithm.util.draw import draw
```

docs/reference/gem.md
```diff
@@ -246,8 +246,9 @@ learner.save('./saved_models/trained_model')
 	This example shows how to perform inference on sample images and draw the resulting bounding boxes using a gem model that is pretrained on the *RGB-Infrared l515 datset*.
 
 ```python
-from opendr.perception.object_detection_2d.gem.gem_learner import GemLearner
-from opendr.perception.object_detection_2d.gem.algorithm.util.draw import draw
+from opendr.perception.object_detection_2d import GemLearner
+from opendr.perception.object_detection_2d import draw
```
jelledouwe, line 250:
```suggestion
from opendr.perception.object_detection_2d.gem.algorithm.util.draw import draw
```

projects/opendr_ws/src/perception/scripts/object_detection_2d_detr.py
```diff
@@ -19,8 +19,9 @@
 import numpy as np
 from vision_msgs.msg import Detection2DArray
 from sensor_msgs.msg import Image as ROS_Image
+from opendr.engine.data import Image
 from opendr_bridge import ROSBridge
-from opendr.perception.object_detection_2d import draw
+from opendr.perception.object_detection_2d import draw_bounding_boxes
```
jelledouwe, line 24:
```suggestion
from opendr.perception.object_detection_2d.detr.algorithm.util.draw import draw
```

projects/opendr_ws/src/perception/scripts/object_detection_2d_gem.py
```diff
@@ -152,8 +152,8 @@ def callback(self, msg_rgb, msg_ir):
         :type msg_ir: sensor_msgs.msg.Image
         """
         # Convert images to OpenDR standard
-        image_rgb = self.bridge.from_ros_image(msg_rgb).numpy()
-        image_ir_raw = self.bridge.from_ros_image(msg_ir).numpy()
+        image_rgb = self.bridge.from_ros_image(msg_rgb).opencv()
+        image_ir_raw = self.bridge.from_ros_image(msg_ir).opencv()
```
jelledouwe, line 156:
```suggestion
        image_ir_raw = self.bridge.from_ros_image(msg_ir, 'bgr8').opencv()
```

docs/reference/detr.md
ad-daniel: You're right. I didn't add it to the `__init__` because there's 2 draw utilities, one for GEM and one for DETR, my bad.

passalis:
@jelledouwe Thanks for catching this. At some point we should probably merge the `draw_bounding_box()` and the `draw()` functions to provide a uniform interface for plotting.

passalis:
> @negarhdr indeed it seems this file no longer is present in the FTP (or at least, not in the owncloud), was it recently removed/renamed?

@negarhdr we restored the old files. Please check them and let us known if any further update is needed.

vniclas:
Please don't merge this yet. The tests for 'efficient_ps' are failing. I didn't see this before but am working on fixing it.

src/opendr/engine/data.py
```diff
@@ -211,15 +211,45 @@ def __str__(self):
...
+            if guess_format:
+                # If channels are found last and image is a color one, assume OpenCV format
+                if data.shape[2] == 3:
+                    data = cv2.cvtColor(data, cv2.COLOR_BGR2RGB)
+                    data = np.transpose(data, (2, 0, 1))
+                # If channels are found last and image is not a color one, just perform transpose
+                elif data.shape[2] < min(data.shape[0], data.shape[1]):
+                    data = np.transpose(data, (2, 0, 1))
+            else:
+                raise ValueError("Image is of type None")
```
vniclas, line 252: @passalis This line has to be moved one level up. I realized this when using the method with `guess_format=False`.

passalis: Thanks for catching another one bug! If pushed a fix for this.

vniclas (dismissed):
As soon as we are sure that all tests pass, we can merge it from my point of view. Thanks a lot @passalis !

passalis:
The tests are broken again due the a change in Image, I will check this in a few minutes and I will let you know when it is ready again.

passalis:
I think I fixed the issue (Image constructor was not called from a subclass, passing the default None argument).

ad-daniel (approved):
Thank you.

vniclas (approved):

PR #165, opened 2021-12-09
Lower ram load

ad-daniel:
I tried a few things but got nowhere, we can restructure the test at a later moment, it's not too important now.

To avoid having too many fails due to memory usage I've lowered limit for parallel runs. It will slow things down a bit, but hopefully it's still faster than having to rerun the tests twice as we have been forced to do lately.

We can rework the test suite when we move to the new repo


ad-daniel:
I've re-run it twice and had no issues so far, it appears to have unlocked the issue [here](https://github.com/tasostefas/opendr_internal/pull/158) as well. It isn't a fully fledged solution but it appears to give us some breathing room at least

PR #166, opened 2021-12-10
single demo grasp + updates based on input consistency

amirmehman21:


ad-daniel:
Hi Amir, is there a reason why you are making the PR from a fork instead of cloning opendr directly and adding your changes to a branch from master ? It's easier for us to add commits if necessary and I don't need to approve the tests manually for them to run.

PS: to fix python formatting errors (automatic `test sources` test) [this](https://github.com/tasostefas/opendr_internal/wiki/Coding-Style#python-coding-style) helps

PS: I tested the previous PR locally and it was working. I'll do the review as soon as I can

ad-daniel:
Also, the CI machines don't have graphic cards, so for the test you need to switch to cpu as done in other tools.

amirmehman21:
Hi

Alright. I'll update it accordingly

Best regards,
Amir
________________________________

ad-daniel:
@amirmehman21 last comment I swear, it's also missing the documentation page for the learner class, to be added under `docs/references` and its contents should be similar in format as done for other tools.
When done, you should also add a reference to this documentation in the `index.md`.

ad-daniel:
Hi @amirmehman21, have you been added as contributor yet?

Just some points that need to be addressed (in the separate PR when you open it).

- I've tested the demo locally and works nicely.
- The unittest doesn't rely on webots, so it might not be a good idea to install webots when running  `install_single_demo_grasp.sh`. It is needed by the demo though, so I would instead just add these instructions directly in the readme of the demo 
- PROTO cache files should not be included in the PR (should be added to a gitignore)
- I haven't been able to successfully run the unittest locally. It attempts to find the GPU still.

As soon as the other PR is up I can help you with some of these points

amirmehman21:
Hi

I am still waiting for the contribution invitation.

About the points:

I will remove webots installation scripts from install_single_demo_grasp.sh and update readme as suggested. Also, I will add proto cache files to git ignore.

In the last PR that I sent, I did not modify tests to use cpu but it is corrected in the version that I’m gonna make the direct PR again. For that, I am still waiting for the invitation.

Best regards,
Amir

ad-daniel:
I think we can close this for now, as we have the other PR open.

PR #168, opened 2021-12-10
EAGERx

jelledouwe:
Dear reviewers,

This PR contains the EAGERx standalone utility framework for reinforcement learning with real and simulated robots. As discussed with Stefania, we have put it in the projects folder. I hope this is the right location. Also, since the code is available on an external repo, we have added EAGERx as a submodule, just like EfficientPS. However, this would mean that we propose to update the [installation instructions](https://github.com/tasostefas/opendr_internal/wiki/Development-System-Setup/) on updating submodules to:
```bash
git submodule update --remote
```
Furthermore, since the toolkit consists of a set of ROS packages we create a catkin_eagerx_ws, just as is done for mobile manipulation.

Finally, the documentation on EAGERx still has to be updated, but we have created this PR before doing so in order to check with you if the structure, location and installation procedure of EAGERx is okay.

Thank you in advance!


projects/control/eagerx/install_eagerx.sh
```diff
@@ -0,0 +1,50 @@
...
+sudo apt-get update && sudo apt-get install -y \
+  ros-${ROS_DISTRO}-ros-base \
+  python3-rosdep || exit;
+source /opt/ros/${ROS_DISTRO}/setup.bash
+
+# Check if EAGERx submodule is initialized
+if [ ! -d ${MODULE_PATH}/eagerx/eagerx_core ]; then
+  STR=$'EAGERx submodule not initialized\nPlease run:\ngit submodule init\ngit submodule update'
+  echo "$STR"
+  exit 1
```
ad-daniel, line 26: It's good to check but this shouldn't happen. In the installation guide we'll need to mention to clone the repo with `git clone --recurse-submodules -j8 https://github.com/...`

If you intend to add a test for this (I don't think so, but if you do), for the `test tools` to work in the CI you'll need to add:

```
- uses: actions/checkout@v2
    with:
       submodules: true
```

to the `tests_sources.yml`, or `submodules: recursive` if it depends on additional submodules.

ad-daniel (commented):
I think the location is fine, being a standalone tool it makes sense to having under projects, likewise the installation procedure is clear and the demo works. I'm not super convinced about the workspace location through, we have some of them just about anywhere at this point. Mobile manipulation and eager are under `lib`, single demo grasp plans on adding one under `projects`, end-to-end planner I'm not sure. @passalis do you know if having it under `lib` was a deliberate choice or if anything else was decided about this?

projects/control/eagerx/install_eagerx.sh
jelledouwe: Thank you for the clarification!

passalis:
> I think the location is fine, being a standalone tool it makes sense to having under projects, likewise the installation procedure is clear and the demo works. I'm not super convinced about the workspace location through, we have some of them just about anywhere at this point. Mobile manipulation and eager are under `lib`, single demo grasp plans on adding one under `projects`, end-to-end planner I'm not sure. @passalis do you know if having it under `lib` was a deliberate choice or if anything else was decided about this?

I think `lib` should be for (compiled) libraries needed to be found during the runtime (i.e. the C OpenDR library ends up there after complilation.)  This is more or less what the current description of this folder reads and I think this is the typical convention. Do we have ROS workspaces in `lib`? I got a bit confused. I think everything is under `project` right now.

ad-daniel:
Yes currently `catkin_ws_mobile_manipulation` is in the `lib` folder, and accordingly in this PR eager would be doing the same. Both should be moved to projects then

passalis:
I missed this... Yes, I agree, they should be under `projects`.

jelledouwe:
> 



> I missed this... Yes, I agree, they should be under `projects`.

@passalis and @ad-daniel, shall I move catkin_ws_eagerx to projects then or should I wait in order to prevent inconsistency?

passalis:
My suggestion is to move it there and then we can open a new PR to fix the other one too.

ad-daniel:
@jelledouwe pep8 test isn't currently passing

ad-daniel:
There seems to be an issue with this PR.
- in the CI, installing compilation dependencies fails, I haven't been able to figure out why just yet. No other PR is having issues with mobile_manipulation, so I assume a change done here should be the cause. I tried locally from a clean state but it doesn't happen, so I'm not sure what's the issue.
- what I can reproduce locally is an issue with conflicting dependencies, I get this:

jelledouwe:
> 

Sorry Daniel, but I do not understand why there are problems with installing the compilation dependencies. I do see the following errors when the tests break:

ERROR: Could not find a version that satisfies the requirement roslaunch (from versions: none)
ERROR: No matching distribution found for roslaunch 
./install_mobile_manipulation.sh: line 37: vcs: command not found
#All required rosdeps installed successfully
update the moveit configs with the global joint
./install_mobile_manipulation.sh: line 51: catkin: command not found
./install_mobile_manipulation.sh: line 53: devel/setup.bash: No such file or directory
make[1]: *** [Makefile:28: install_compilation_dependencies] Error 1
make: *** [Makefile:40: install_compilation_dependencies] Error 2

ad-daniel:
Seems to be working now, I changed the demo instructions as they were not working.
Please double-check the changes, otherwise it looks good to me.

passalis (approved):
Thank you!

docs/reference/index.md
```diff
@@ -65,6 +65,8 @@ Neither the copyright holder nor any applicable licensor will be liable for any
         - [mobile_manipulation Module](mobile-manipulation.md)
     - `simulation` Module
         - [human_model_generation Module](human-model-generation.md)
+    - `Stand-alone Utility Frameworks`
+        - [eagerx](eagerx.md)
```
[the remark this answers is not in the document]
jelledouwe: Thanks for the suggestion, I will change it!

jelledouwe:
> Seems to be working now, I changed the demo instructions as they were not working. Please double-check the changes, otherwise it looks good to me.

Thank you very much Daniel. I checked the changes and made some minor changes afterwards. Running the roscore was not necessary, so I removed it from the instructions to make them as simple as possible.

docs/reference/index.md
```diff
@@ -55,7 +55,7 @@ Neither the copyright holder nor any applicable licensor will be liable for any
             - [multilinear_compressive_learning Module](multilinear-compressive-learning.md)
         - semantic segmentation:
             - [semantic_segmentation Module](semantic-segmentation.md)
-        - panoptic segmentation:
+        - panoptic segmentation:s
```
ad-daniel, line 58:
```suggestion
        - panoptic segmentation:
```

ad-daniel (changes requested):

ad-daniel (approved):
Thank you!

passalis (approved):

PR #175, opened 2021-12-14
ROS nodes for speech command recognition

anton-muravev:
This PR contains ROS nodes for speech command recognition models: MatchboxNet and Quadratic SelfONN. As mentioned before, EdgeSpeechNets are not supported at this time, as there is no pretrained model available.

These ROS nodes make use of AudioData messages, which are provided by _audio_common_msgs_ ROS package. The dependencies are updated accordingly.

Note that I've chosen to not implement a function for converting between ROS and OpenDR data types as part of the bridge. Audio input needs to be accumulated for a certain duration (given as a user-defined parameter) to be able to be recognized. I currently do this by appending the incoming data to the Numpy array in the node script. When the desired length is reached, this array is converted to the OpenDR Timeseries object. It seems redundant to make a one-line function in the bridge for this which would only call the standard Timeseries constructor (no ROS messages involved). On the other hand, moving the input accumulation logic into the bridge seems to go beyond its scope. Of course if we decide it's best to have a function in the bridge after all, I will make the required changes.


passalis (approved):
Thank you! Seems fine to me!

ad-daniel (approved):
Thank you!

PR #183, opened 2021-12-15
Semantic segmentation demos

mtzelepi:
Semantic segmentation demos


ad-daniel (dismissed):
I've fixed the demos since they weren't running due to the wrong imports, thank you!

passalis (approved):
Thank you! I added some code to visualize the inference results.

ad-daniel (approved):
Thank you!

PR #187, opened 2021-12-18
ROS nodes for speech command recognition, extension

anton-muravev:
I modified the previously merged ROS node to allow for an option to load the pretrained model locally instead of downloading it from the server. This is slightly more general, but more importantly, allows the node to support EdgeSpeechNets as well.


passalis (approved):
Thank you!

ad-daniel (approved):

PR #188, opened 2021-12-18
Missing demo for speech command recognition

anton-muravev:
A demo script to show how the speech command recognition can be performed on a local audio file. Some examples from the public Google Speech Commands dataset are included.


ad-daniel:
`matchboxnet` and `quad_selfonn` don't work for me (cuda related issues) and `edgespeechnets` throws a "No pretrained EdgeSpeechNets model available for download`. I've tried re-running the unittests of these learners and they work just fine. Is it just my system?

anton-muravev:
@ad-daniel 
Regarding Edgespeechnets - this is related to the fact that there is no pretrained model for this method, which came up also recently in the context of ROS nodes. So, as README describes, you do need to have the model trained locally and pass an argument with the path to the model. If this path is not given, the script attempts to download the pretrained model from FTP, which is valid for MatchboxNet and Quadratic SelfONN, but not EdgeSpeechNets. So this result is intended.

Regarding cuda issues - this result is not intended... Any details you can give on these, if they are consistent? Not sure what could go wrong given the simplicity of the script.

passalis:
For me MatchboxNet and Quadratic SelfONN seem to work ok both on cpu and gpu.

ad-daniel (approved):
Cuda problems were on my side, been having issues with it all day. They both work now however. Thank you!

passalis (approved):

PR #191, opened 2021-12-26
Prerelease bugfixes

passalis:
This PR contains some minor bugfixes:
1. Fixed wrong LD_LIBRARY_PATH definition on `install.sh`
2. Added installation for usb cam dependencies (broken on docker installations)
3. Fixed a typo and usage instructions for anomaly detection demo

### DIFF
--- a/bin/activate.sh
+++ b/bin/activate.sh
@@ -2,7 +2,7 @@
 export OPENDR_HOME=$PWD
 export PYTHONPATH=$OPENDR_HOME/src:$PYTHONPATH
 export PYTHON=python3
-export LD_LIBRARY_PATH=$OPENDR_HOME/src:$LD_LIBRARY_PATH
+export LD_LIBRARY_PATH=$OPENDR_HOME/lib:$LD_LIBRARY_PATH
 
 if [[ -z "${OPENDR_DEVICE}" ]]; then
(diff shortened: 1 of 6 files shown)


passalis:
This PR is now ready for review. It only contains minor fixes (mainly typos), apart from two 2 minor code issues (the first was affecting C API, since `LD_LIBRARY_PATH` was not set correctly, and the second one the installation of ROS on docker containers).

iliiliiliili:
I have also updated setup.py for 3D detection and 2D tracking demos.

ad-daniel (approved):
Nice catch, thank you.
Github actions on the other hand seems to be acting up. I don't think anything changed on our end, and this PR was passing yesterday, today I've manually re-run it (end-to-end planner was having it too) and it fails to checkout the submodules. It only appears to affect the CI though, don't have problems with it locally.

Pavlos-Tosidis (approved):

ad-daniel:
6th time is the charm apparently

ad-daniel:
github actions are having problems as a whole: https://www.githubstatus.com/
we can bypass it and merge faster if needed, but usually they fix these issues fairly quickly

passalis:
I think we can wait until tomorrow and if the problem remains, then we can bypass the checks.

PR #198, opened 2022-01-12
upgrade scikit-learn to 0.22

Pavlos-Tosidis:
0.21.3 causes warnings with the current numpy version. Upgrading it fixes it.


ad-daniel (approved):
Looks good to me, I don't see any side-effects in the installation.
However we should document these changes going forward. I've added a change log.

PR #201, opened 2022-01-20
Modular python packages

passalis:
It would be good if we were able to split OpenDR into subpackages that can be individually installed. We already have the `dependencies.ini` that can be used for finding out the dependencies. Then, the `opendr-toolkit` package could act as meta-package that ensures that all the individual packages are installed (for users that want to install the whole toolkit), while individual tools could be also installed. For example, for face recognition a user could only run ` pip install opendr-toolkit-face-recognition`, which will only install face recognition, while it should all pull `opendr-engine` which will contain the base OpenDR functionality shared between different tools.

Tasks:
- [x] Switch into platform agnostic `sdist` format
- [x] Adjust `setup.py` for building separate packages
- [x] Ensure that `opendr-toolkit-engine` contains all the shared functionality
- [x] Create an appropriate metapackage
- [x]  Create appropriate documentation (e.g., for using both with CPU and GPUs and jetsons)
- [x] Update changelog
- [x] Update github actions/testing pipeline

Tools included so far:
- [x] engine/utils
- [x] perception/speech_recognition
- [x] perception/semantic_segmentation
- [x] perception/face_recognition
- [x] perception/pose_estimation
- [x] utils/hyperparameter_tuner
- [x] perception/compressive_learning
- [x] perception/facial_expression_recognition
- [x] perception/heart_anomaly_detection
- [x] perception/multimodal_human_centric
- [x] perception/activity_recognition
- [x] perception/skeleton_based_action_recognition
- [x] simulation/human_model_generation
- [x]  perception/object_detection_2d
- [x] perception/object_detection_3d
- [x] perception/object_tracking_2d
- [x] perception/object_tracking_3d
- [x] perception/panoptic_segmentation
- [x] control/single_demo_grasp
- [ ] ~~control/mobile_manipulation~~

Update:
This will probably work. I have automated the generation of different packages under the same namespace. I have also included as post-install scripts the installation of necessary dependencies from git repositories. I think after having this completed we will manage to:
- simplify the installation of pip packages (since we will no longer need to manually install the packages from github)
- allow for removing the hard dependency on CUDA10.2 and allow the pip packages to work both on cpu and gpu, since this arises only from two packages which now defaults to cpu. The user can install cuda-enabled versions of mxnet and detectron to enable gpu acceleration for the tools that need these packages
- potentially support a very wide range of python versions due to relaxing the required versions
-  greatly accelerate installation speed


passalis:
This is in good shape. I have managed to validate that all tools runs correctly apart from mobile_manipulation which requires a functional ROS installation (it didn't work with the previous package either). I will wait for #208 to be merged, in order to also incorporate these packages into the github action pipelines.

passalis:
I think this is ready for review if the tests pass. I have also added another set of tests where each tool is tested when only its package (along with 3 other dependency packages) is installed (instead of the whole toolkit).

passalis:
It seems that it was indeed a good idea to test the wheels individually, there are some missing dependencies that are revealed only when the packages are installed separately. I am fixing them.

ad-daniel (commented):
Looks great to me, thank you!
I believe it only misses updating the develop workflow of test_suite (same thing, but refers to the develop branch). 
We also need to update the test_packages workflow but only after the release, since it pulls them from pypi.
Concerning end-to-end planner at this point the PR should be recreated and target develop and do the necessary adaptations there. I think we should prioritize merging this one first

passalis:
Thanks!
I think I did it. I added wheel testing to the `tests_suite_develop.yml`. Please have a quick look and let me know if you think that needs any update.

However, now I see a fail on cppcheck on macos without changing anything :-/ I will also fix this.

ad-daniel (approved):
Looks fine to me, only made some cleanup, thank you!
Yes, that's normal as newer versions of cppcheck are released (sometimes on windows, other times on linux/mac) it detects/complains about different things. So every now and then it requires fixing or suppressing what it suggests.

Pavlos-Tosidis (approved):
Thank you Nikos. Great work!

PR #204, opened 2022-01-27
Automatic local build/install/test of wheel from a branch

ad-daniel:
The wheel is first built, then sent to another job for it to be installed and tested. The `test release` label is meant to be used especially before an actual release takes place (before creating a tag), this ensures the wheel that will be pushed to [pypi.org](pypi.org) will work as intended, without having to wait for the nightly test to find out.

### DIFF
--- a/.github/workflows/publisher.yml
+++ b/.github/workflows/publisher.yml
@@ -1,0 +1,72 @@
+name: Publisher
+
+# Trigger on new github release, a tag with format vX.Y.Z is expected (used to tag the docker)
+on:
+  release:
+    types: [published]
+
+env:
+  OPENDR_VERSION: ${{ github.event.release.tag_name }}
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  publish-wheel:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install prerequisites
+      run: |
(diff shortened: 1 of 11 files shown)


passalis (dismissed):
Thank you! This is very nice!

I have only one concern/question: Is there a limit on the total size of artifacts? If so, should we reduce the retention period (in the settings) from 90 days to something smaller or the older ones are automatically deleted if needed?

ad-daniel:
To my knowledge, github automatically purges older artifacts until at most you're using a total amount of space at any given time (not sure what that amount is, but I think is 10GB). There's no way of deleting artifacts other than doing so manually. In practice however you are right, we don't really need to retain them nor do we need 90 days worth of logs, so we can lower the retention period. A week for our purpose is more than enough. Either way for our purpose, the minimal retention of 1 day suffices, so I enforced this now in the workflow.

Under `account > settings > billing and plan > switch to organization account` shows the current monthly consumption for our current plan (free). Maybe I don't have the full access to it, but when I check everything is at zero (or it just reset). We are nowhere near maxing it out on Webots, I'd be surprised if we are in OpenDR

ad-daniel:
Actually never mind:
> GitHub Actions usage is free for both public repositories and self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage, depending on the product used with the account.

The reason usage is zero is because we have a public repo. The limits mentioned in that page count only towards private repos, which we don't have so we're fine.

passalis (approved):

ad-daniel:
I don't know why, but it seems messing with the artifact retention date ended up breaking this PR, rerunning doesn't solve it. I try to re-open it.

PR #207, opened 2022-01-31
Fix 3D Detection And 2D Tracking Tests

iliiliiliili:
Fix for 3D object detection `free pointer error`


iliiliiliili:
@passalis @ad-daniel It does not seem to appear in this PR, but I removed `del learner` lines that are most likely to cause this. It was used to reduce RAM load when all tests were done inside a single test group.

iliiliiliili:
This error persists for 3D tracking tests too and most likely is a problem if g++ compilation and python bounding

iliiliiliili:
I have resolved a problem with slow data loader for KITTI and simplified testing models and tests, so the memory issue should have much fewer chances to occur.

stefaniapedrazzi:
This PR contains many unrelated changes.
It may be that it is based on the `master` branch but targeting the `develop` branch.

If the test error is also reproducible on the master branch, then this PR should be merged to the master branch and not the develop branch.

ad-daniel:
I've run this version multiple times and it seems that both `object_detection_3d` and `object_tracking_3d` are working now, it hasn't run out of memory so far either, so the simplification seems effective.
If you think the current version of the test is representative of the tool and that what we test suffices to notify us if future changes (like changing dependencies to some other tool) breaks this one then I think that after a final cleanup (removing commented lines) we can proceed to merge the PR.

ad-daniel (approved):
Thank you!

passalis (approved):
Thank you!

PR #208, opened 2022-02-01
Automatic local build/install/test of wheel from a branch

ad-daniel:
The wheel is first built, then sent to another job for it to be installed and tested. The test release label is meant to be used especially before an actual release takes place (before creating a tag), this ensures the wheel that will be pushed to pypi.org will work as intended, without having to wait for the nightly test to find out.

### DIFF
--- a/.github/workflows/tests_suite.yml
+++ b/.github/workflows/tests_suite.yml
@@ -12,7 +12,7 @@ defaults:
 
 jobs:
   cleanup-runs:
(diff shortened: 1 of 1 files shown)


ad-daniel:
Looks like whatever was causing issues with the artifacts is gone, the PR can now be reviewed

passalis (approved):
Thanks!

passalis:
I think the failures are caused on the github end (e.g., this seems like a connectivity issue https://github.com/opendr-eu/opendr/runs/5035624304?check_suite_focus=true#step:3:181)

PR #211, opened 2022-02-03
Merge `master` into `develop`

passalis:
Synchronizes the develop branch with the latest changes from master.

### DIFF
--- a/.github/workflows/publisher.yml
+++ b/.github/workflows/publisher.yml
@@ -1,0 +1,72 @@
+name: Publisher
+
+# Trigger on new github release, a tag with format vX.Y.Z is expected (used to tag the docker)
+on:
+  release:
+    types: [published]
+
+env:
+  OPENDR_VERSION: ${{ github.event.release.tag_name }}
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  publish-wheel:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install prerequisites
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
+    - name: Build Wheel
+      run: |
+        ./bin/build_wheel.sh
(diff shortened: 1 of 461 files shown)


passalis:
I created #212 for resolving these conflicts from another branch. @ad-daniel is this correct or is there an easier way of doing this?

ad-daniel:
If there are conflicts it's usually easier to do the synchronization manually:

0. pull latest changes from master into master, and from develop into develop
1. create a branch from develop called something like: `merge-master-into-develop`
2. merge master into it and solve any conflict
3. push and open PR

ad-daniel:
There's actually a step 4, when merging the PR you shouldn't do "Squash and merge" but actually do "Merge commit" to keep the linearity. To do so however requires temporarily remove the restriction (in this case from develop branch), by going settings > branches > edit develop protection rules > disable  "Include administrators"

After the merge, you need to re-activate the protection

passalis:
I think this is what I did, but I will check again later since the number of changed files do not agree between the PRs, which doesn't sound right  :stuck_out_tongue:



<br class="Apple-interchange-newline">

passalis:
Tbh I cannot understand why this PR shows so many changed files. For example, `MANIFEST.in` appears as a new file in this [PR](https://github.com/opendr-eu/opendr/pull/211/files#diff-41d5a52589e0480be9c099d2bba7a8135b8b0d71bcbb8df3582a8df1c2295003) even though it is not: [master](https://github.com/opendr-eu/opendr/blob/master/MANIFEST.in) vs [develop](https://github.com/opendr-eu/opendr/blob/develop/MANIFEST.in)

For #213 I have done the following (which was more or less the same as before):
```bash
git checkout develop
git checkout -b merge-master-into-develop develop 
git merge --no-ff master
--- resolved the conflicts ---
git add .github/workflows/*
git add -A
git commit -m "Conflict resolution"
git push
```

ad-daniel:
Most of the files are about the change in changelog dates, which in principle has been synchronized into the develop branch already (I think I did it last week). The develop branch also has the file you mentioned [here](https://github.com/opendr-eu/opendr/blob/develop/MANIFEST.in), so no clue why it adds it as a new file despite existing.
Either way the other PR you just made looks ok

PR #215, opened 2022-02-08
Upgrade to CUDA 11.2 and improve GPU support

passalis:
This PR upgrades the toolkit to using CUDA 11.2. This also ensures that the toolkit will be compatible with the NVIDIA 30xx GPUs. For PyTorch we are using precompiled packages that bundle CUDA11.1. This does not affect the system-wide CUDA version.

This PR also improves testing using GPUs, as well as fixes some documentation issues regarding the use of `OPENDR_DEVICE` variable.

Tasks to be performed
- [x] Change dockerfile to using CUDA 11.2
- [x] Update PyTorch and mxnet
- [x] Update detectron
- [x] Update DCNv2
- [x] Make sure that pip installation does not need any kind of update
- [x] Update the documentation if needed

We need to restore github branch in Dockerfile prior to merging.


passalis:
I think this is ready for review. Artifact download seems to randomly fail for some tests (even though its works fine for others... :-/ ). I hope that this run will complete correctly.

ad-daniel:
I took the liberty of re-synchronizing this branch after the merges and it appears to be working CI-wise, so the change doesn't break anything, but of course the CI doesn't have GPUs so it's the local test that matters here. Haven't checked yet.

The only doubt I had was concerning the platforms such as TX2, I was under the impression they need lower cuda versions, does this change affect them and will we need to provide a separate images/packages for them?

passalis:
Actually I have found an issue with mxnet installation. I putting to this to draft again to fix this.

passalis:
I think that I have fixed the issues by installing detectron from source (it is quite fast and it is linked against the currently installed pytorch). We will test this manually on GPU to make sure that it still works as expected. Otherwise, I think (and hope) that this is ready for review.

passalis:
After some testing we found out that numba 0.48 was incompatible with CUDA11.2. I have upgraded numba to 0.53.

passalis:
I also removed the reference to `OPENDR_DEVICE` from `activate.sh` since this variable is never used after installation. We only use it to install the appropriate dependencies during the installation. The inference device is always selected by the user when the toolkit is used from the Python API.

passalis:
Object detection tests are now going out of memory (probably due to upgrading numba). I have separated face detection tests from the object detection tests to reduce memory consumption.

@iliiliiliili I have added a few statements for placing the model to the correct device, since it seems that it moved after a few operators, e.g., training. I cannot understand this behavior, but tests weren't able to run without them when using GPU acceleration. Please have a look and let us know if this is not correct.

passalis:
I do not want to be too excited but I think that I solved it unless something else fails now... :-D We are going to recheck all tools and reported back if they pass ok.

passalis:
Panoptic segmentation is now broken due to upgrading PyTorch from 1.8. to 1.9 ...
The only solution I am seeing right now is downgrade to Pytorch 1.8, which will then break detectron. So then we will have to switch to CUDA 11.3 which works on detectron. But this will then break mxnet, for which I couldn't find any package that supports 11.2. 

@vniclas Is there any possibility that this an easy fix for making panoptic segmentation work with PyTorch 1.9? I am getting this error https://github.com/qubvel/segmentation_models.pytorch/issues/424 and https://github.com/NVIDIA/apex/pull/1049

@vivinousi Is there any possibility that mxnet packages for 11.2 exist? What other options do we have?

vivinousi:
> Panoptic segmentation is now broken due to upgrading PyTorch from 1.8. to 1.9 ... The only solution I am seeing right now is downgrade to Pytorch 1.8, which will then break detectron. So then we will have to switch to CUDA 11.3 which works on detectron. But this will then break mxnet, for which I couldn't find any package that supports 11.2.
> 
> @vniclas Is there any possibility that there is an easy fix for making panoptic segmentation work with PyTorch 1.9? I am getting this error [qubvel/segmentation_models.pytorch#424](https://github.com/qubvel/segmentation_models.pytorch/issues/424) and [NVIDIA/apex#1049](https://github.com/NVIDIA/apex/pull/1049)
> 
> @vivinousi Is there any possibility that mxnet packages for 11.2 exist? What other CUDA options do we have for mxnet?

I haven't found any prebuilt whl files for mxnet with cuda 11.2. We could try to build it ourselves from source and provide the whl?

passalis:
@vivinousi  I meant CUDA11.3... Yes, perhaps this could be an option.

vniclas:
@passalis I'll try to check this someday this week if I have time. I'll be on vacation next week though, i.e., I might only be able to look into this after March 28. It would definitely be faster if you just try the solutions proposed in the threads linked by you. If they make the tests pass again, I can propagate those changes to the original repository.

vniclas:
@passalis I fixed it in the linked repository. Please try again after having re-cloned and re-installed it.

passalis:
@vniclas thank you! I can confirm that this is working now.

docs/reference/installation.md
```diff
@@ -36,9 +36,9 @@ Using dockerfiles is strongly advised (please see below), unless you know what y
 Please also make sure that you have enough RAM available for the installation (about 4GB of free RAM is needed for the full installation/compilation).
 
 
-You can set the inference/training device using the `OPENDR_DEVICE` variable.
-The toolkit defaults to using CPU.
-If you want to use GPU, please set this variable accordingly:
+If you want to install GPU-related dependencies, then you can appropriately set the `OPENDR_DEVICE` variable.
+The toolkit defaults to using CPU.
```
ad-daniel, line 40:
```suggestion
The toolkit defaults to using CPU.
```

docs/reference/installation.md
```diff
@@ -49,24 +49,20 @@ source ./bin/activate.sh
...
-sudo apt install gcc-8 g++-8 gcc-9 g++-9
-sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 8 --slave /usr/bin/g++ g++ /usr/bin/g++-8
-sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 9 --slave /usr/bin/g++ g++ /usr/bin/g++-9
-echo "Please switch to GCC 8"
-sudo update-alternatives --config gcc
-```
-Then, you can install CUDA, along CuDNN.
-You can also refer to this [dockerfile](https://github.com/opendr-eu/opendr/blob/master/Dockerfile-cuda) for installation instructions.
-Note that NVIDIA 30xx GPUs may not be fully supported, due to CUDA limitations.
+If you plan to use GPU-enabled functionalities, then you are advised to install [CUDA 11.1](https://developer.nvidia.com/cuda-11.1.0-download-archive), along with [CuDNN](https://developer.nvidia.com/cudnn).
```
ad-daniel, line 62:
```suggestion
If you plan to use GPU-enabled functionalities, then you are advised to install [CUDA 11.2](https://developer.nvidia.com/cuda-11.2.0-download-archive), along with [CuDNN](https://developer.nvidia.com/cudnn).
```

ad-daniel (changes requested):
Thanks, only 2 details.
I'll begin testing now

passalis:
Thanks for the extensive tests!!!

I have reproduced the errors. I am looking into them.

passalis:
Regarding the `facial expression recognition` & `skeleton_based_action_recognition`: This is either a bug in the way ONNX handles the conversion or part of the model stays in CPU. I followed the workaround proposed in PyTorch community (i.e., temporarily move the model to CPU only for the conversion process) and it seems that it works.

Regarding  `cox3d.test_cox3d_learner.TestCoX3DLearner`:
I managed to replicated the issue on a 3080 but not on a 1060 Ti. So I assume that this is not a bug in our code. However, the code works correctly if we pass a batch of samples instead of single sample for inference. This sounds like a bug, but on the other hand it still works correctly on CPU and on some GPUs. So, I have applied this change, which - however - just masks the issue. @LukasHedegaard could you have a look to make sure that this no bug on our code that can cause this behavior? Better to use a 30xx GPU for testing, otherwise you might not be able to replicate this bug.


> Another weird thing is that I haven't been able to run the jupyter notebooks with GPU. If I run the demo `inference_demo.py` it works fine with the GPU, but running `inference_tutorial.ipynb` I get:
> `RuntimeError: CUDA error: no kernel image is available for execution on the device`

For which demo this happens? Can you verify that you are using `jupyter` from the venv instead of a globally installed version?

LukasHedegaard:
> Regarding cox3d.test_cox3d_learner.TestCoX3DLearner: I managed to replicated the issue on a 3080 but not on a 1060 Ti. So I assume that this is not a bug in our code. However, the code works correctly if we pass a batch of samples instead of single sample for inference. This sounds like a bug, but on the other hand it still works correctly on CPU and on some GPUs. So, I have applied this change, which - however - just masks the issue. @LukasHedegaard could you have a look to make sure that this no bug on our code that can cause this behavior? Better to use a 30xx GPU for testing, otherwise you might not be able to replicate this bug.
It is likely that the numerical precision may differ a bit between hardware platforms. In the quoted line (`test_cox3d_learner.py, line 115`) a list of opendr.Image is converted to a batch inside the infer function (`torch.stack([torch.tensor(v.data) for v in batch])`) and passed on. The difference in precision must be due to the conversion between torch.Tensor and opendr.Image (internally np.array). Unfortunately I do not have a 30xx GPU at hand for testing, but I am fairly confident that a reduction of the tolerance (e.g. atol=1e-5 or atol=1e-4 in `test_cox3d_learner.py line 115` ) should fix the assertion-error. 4th decimal differences do not matter on this case.

ad-daniel:
I've tested again and `facial expression recognition` & `skeleton_based_action_recognition` are indeed working now for me as well.

I've also tested `cox3d.test_cox3d_learner.TestCoX3DLearner` on the 3060 and relaxing the tolerance does indeed work for me, thanks for looking into it!

> For which demo this happens? Can you verify that you are using jupyter from the venv instead of a globally installed version?

Yes you're right, my bad, it was using jupyter installed from apt. When using from venv it does indeed work, thanks!

ad-daniel:
I believe everything works on the repo side now, but we should probably test the docker as well and if that works we can merge (building mine now).

ad-daniel:
Does the cuda docker work for you? Despite activating the venv several dependencies appear to be missing in mine (opencv, onnxruntime, sklearn, bcolz). If that were the case however the cpu docker should fail as well in the CI but doesn't appear to. Is it just me?

passalis:
We have tested the GPU docker before the last commits and it worked. I will check again and let you know.

ad-daniel:
No luck on my end, still appears to be missing the dependencies. I'll try on a different PC just in case

passalis:
Is there any error reported during building? This is usually caused by a failed to install dependency which in turn causes the whole install script to exit.

ad-daniel:
Not that I noticed, but I'll check now that I rebuild it on the laptop

passalis:
I realized that CPU dockerfile didn't use this branch when building, so I fixed this (we still need to revert this when merging this PR). So, I have updated the cpu dockerfile to make sure that the tests use this branch. We are still looking into replicating the issue with the missing dependencies (I have also checked on another machine (2070 Super) and it seems that at least the installation worked - didn't get any failed imports - however we also have other people testing on different machines and configurations).

ad-daniel:
Well, the CI docker tests passed regardless, so it might just mean it's an issue on my end, although I don't know why.
So you are able for instance of running the `inference_demo.py` of face_recognition without any issue with the imports?
If you still have the image, perhaps I can try running yours instead (if you upload it to dockerhub or something).

passalis:
I tried this one and it seems to works. I am uploading the image to https://hub.docker.com/r/passalis/opendrdraft

ad-daniel:
I can confirm the demo works on your image. In fact, when I pulled your image it recognized parts of it and marked them as "already known" and downloaded the rest, so I think it really means it was borked/incomplete on my end. It's still unclear to me why, but for the time being I can test your image so that we can merge as soon as possible

ad-daniel:
All tests worked on my machine, I've also checked a couple random demo and notebooks and they too work as intended.
So as far as I'm concerned we can merge this PR (after removing the branch from the docker files)

ad-daniel (approved):

ManosMpampis (approved):

PR #224, opened 2022-02-14
XML dataset addition + mxnet-cu102 fix

vivinousi:
This PR adds XML-based dataset compatibility with OpenDR format. This includes any dataset in VOC .xml format, like those made with the labelImg tool.
The PR also fixes a dependency issue with mxnet-cu102, where mxnet==1.8.0 was installed after the cuda version and overrode the cu102 installation.


passalis (approved):
Thank you! Seems fine to me. Should we also add a brief reference in the documentation regarding the new DetectionDataset?

passalis (approved):
Thank you!

ad-daniel (approved):
Thank you!

PR #226, opened 2022-02-17
Gem Mobilenet-v2 pretrained backbone

jelledouwe:
Add pretrained mobilenet v2 backbone for the GemLearner and update docs + test.


passalis (approved):
Thank you!

PR #227, opened 2022-02-18
Fix failing nightly test on master

ad-daniel:
Although it appears to work fine for develop, the nightly tests on the master branch are still failing due to messing with the artifact retain duration some time back. Perhaps forcing a delete every time instead might help


omichel (approved):
Looks good to me.
Thank you.

PR #228, opened 2022-02-20
EAGERx as dependency instead of submodule.

jelledouwe:
Dear reviewers,

In order to ease the installation of EAGERx, we converted it from a ROS package to a pure python package. Therefore, it can be installed using pip and does not require to build work spaces. This PR contains updates of the docs and demos that are related to this conversion from a ROS package to a pure Python package. Also, some dependency clashes had to be resolved.


passalis:
Hi, this seems fine to me. My only concern is to ensure that the changes in the dependencies  do not break the pip installation (most probably will be fine, but just to be on the safe side). I have enables `test packages` which tests this. If it fails, I will try to test it locally.

ad-daniel:
To test packages build from this branch we should run `test release`, `test packages` checks the docker image/wheels currently hosted at dockerhub/pypi.
Since the latter failed however, it means we might have an issue in the wheels we are currently distributing.

ad-daniel (dismissed):
Looks good to me thank you. Perhaps I'm remembering wrong but didn't the demo use to visualize the pendulum graphically? The current ones don't, I'm just wondering if it's intended or if something isn't working correctly (or equally likely, if I'm remembering wrong)

jelledouwe:
> Looks good to me thank you. Perhaps I'm remembering wrong but didn't the demo use to visualize the pendulum graphically? The current ones don't, I'm just wondering if it's intended or if something isn't working correctly (or equally likely, if I'm remembering wrong)

Yes you are right, rendering was performed before but we turned it off per default now. But I think it is nice to have the option to easily render the output, so I have added an argument --render to the demos to enable rendering. Thank you!

ad-daniel (approved):
Nice, thank you!

passalis (approved):
Thank you!

PR #233, opened 2022-02-28
Fix master workflow related to artifacts

ad-daniel:
Let's start to see if the issue on the master artifact can be solved, and if so then we'll propagate and adapt the develop one


omichel (approved):

passalis (approved):
Thank you!

PR #240, opened 2022-03-08
Test develop branch

ad-daniel:
Sanity check of develop branch


ad-daniel:
Haven't been able to reproduce the issue that we have in the nightly-develop test, as it appears to usually involve object_detection it might fix itself when merging the change PR about cuda 11, since these tests are split further. Can close for now.

PR #202, opened 2022-01-21
Add automatic publishing of wheel and docker

ad-daniel:
This PR adds automatic build and publishing of python wheel and docker images.

https://github.com/opendr-eu/opendr/pull/200 needs to be merged first.

Just for testing at this moment.

PR #242, opened 2022-03-10
Lightweight openpose fixes

### LINKED ISSUE #36: Lightweight OpenPose optimize fails to run if num_refinement_stages != 2
On [this](https://github.com/tasostefas/opendr_internal/blob/2f6d15fab31ddb3c604f0bd6a2a7e8689904e051/src/perception/pose_estimation/lightweight_open_pose/lightweight_open_pose_learner.py#L847) line, the naming of the outputs requires `num_refinement_stages` to be set to the default value of 2, setting it to another value makes the `optimze` method to fail to run.

[Here](https://github.com/tasostefas/opendr_internal/blob/2f6d15fab31ddb3c604f0bd6a2a7e8689904e051/src/perception/pose_estimation/lightweight_open_pose/lightweight_open_pose_learner.py#L844) the names need to be set dynamically according to `num_refinement_stages`.

tsampazk:
- Fixed height check on transformations.py according to [original lightweight open pose repo #202.](https://github.com/Daniil-Osokin/lightweight-human-pose-estimation.pytorch/pull/202)
- Fixes #36 
- Fixes a similar bug to the above, when using half precision and then trying to optimize the model with onnx

(Please add any appropriate tags missing)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ Released on XX, XXth, 2022.
     - `panoptic_segmentation/efficient_ps`: updated dataset preparation scripts to create correct validation ground truth ([#221](https://github.com/opendr-eu/opendr/pull/221)).
     - `panoptic_segmentation/efficient_ps`: added specific configuration files for the provided pretrained models ([#221](https://github.com/opendr-eu/opendr/pull/221)).
     - `c_api/face_recognition`: pass key by const reference in `json_get_key_string()` ([#221](https://github.com/opendr-eu/opendr/pull/221)).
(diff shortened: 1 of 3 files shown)


ad-daniel (dismissed):
Thank you!
I think it warrants a mention in the changelog.

tsampazk:
> Thank you! I think it warrants a mention in the changelog.

I entirely missed the changelog! Should i add something similar to this PR's description in three bullets in the bugfixes section of the changelog and commit it through this PR?

ad-daniel:
You should add just a short summary (1 phrase or 2 maximum) and at the end link this PR itself as is done on the other examples present there already. If people need more details they can use that link to reach this page and read more about it

Yes, you add the change to the CHANGELOG file and commit it in this same PR

passalis (approved):
Thank you!

ad-daniel (approved):
Thank you!

PR #250, opened 2022-04-25
Fix clang on macOS

ad-daniel:
We need the fix on the master side as well or the nightly tests will continue to fail because of it.


omichel (approved):

passalis (approved):
Thanks!

PR #195, opened 2022-01-10
Tx2 install

Pavlos-Tosidis:
TX-2 installation script.
Updated installation.md and fixed some import statements in object_detection_2d/detr and object_detection_2d/gem

PR #255, opened 2022-05-09
Fix nvidia docker key

ad-daniel:
As reported by Thomas, the docker build process should fail due to outdated nvidia keys, as mentioned [here](https://forums.developer.nvidia.com/t/notice-cuda-linux-repository-key-rotation/212772). This PR addresses the issue, and in practice I assume this PR https://github.com/opendr-eu/opendr/pull/195 might need to do something similar to the Dockerfile-embedded.


ad-daniel:
It seems we cannot remove the previous keys just yet since it appears to still be needed by some of the repos included in base image. Other than that, the gpu-enabled docker image builds on my machine after the key fix

omichel (approved):

Issue #259, opened 2022-06-01
Skeleton-based Human Action Recognition Demo FAILS


davelbit:
After cloning the repository I was keen on testing the skeleton based action recognition, by issuing the following command:
`python3 demo.py --device cuda --video 0 --method stgcn --action_checkpoint_name stgcn_ntu_cv_lw_openpose` as suggested here: 
`https://github.com/opendr-eu/opendr/blob/master/projects/perception/skeleton_based_action_recognition/REAMDE.md`

But the program ended with the following error message:
`Traceback (most recent call last):
  File "demo.py", line 209, in <module>
    category_labels = preds2label(prediction.confidence)
  File "demo.py", line 112, in preds2label
    class_scores, class_inds = torch.topk(confidence, k=k)
TypeError: topk(): argument 'input' (position 1) must be Tensor, not float`

passalis:
Hi davelbit,

Thanks for your interest in our toolkit and for pointing out this bug. We replicated this bug and we are looking into this. We will get back soon.

passalis:
We have identified the cause  and fixed this issue in #260. Could you try the [HAR_demo_fix](https://github.com/opendr-eu/opendr/tree/HAR_demo_fix) branch to see if this fixes the issue for you?

davelbit:
I have cloned the HAR_demo_fix branch as suggested but found that the error still seems to exist. Can you confirm that the error is still present on your end?

passalis:
The bug related to `TypeError` seems to have been resolved for me. Please make sure that you source the new environment after cloning the branch (e.g., make sure that `PYTHONPATH` is pointing to the corrected clone). Let us know if this fixes the issue.

davelbit:
I missed to source the environment. It works now for me as well.

PR #260, opened 2022-06-02
skeleton-based HAR demo bug fixed

negarhdr:
This PR fixes the bug of the skeleton_based_human_action_recognition demo, which is mentioned in #259 yesterday. 
The demo used to work before, and this bug is created due to the last changes that I've mistakenly made in the learners recently. Now it should work without any problem. It's better to double-check. 

The branch is on top of the master branch!

### DIFF
--- a/projects/perception/skeleton_based_action_recognition/demos/demo.py
+++ b/projects/perception/skeleton_based_action_recognition/demos/demo.py
@@ -205,10 +205,11 @@ if __name__ == '__main__':
             if counter > 0:
                 skeleton_seq = pose2numpy(args, counter, poses_list)
 
-            prediction = action_classifier.infer(skeleton_seq)
-            category_labels = preds2label(prediction.confidence)
-            print(category_labels)
-            draw_preds(img, category_labels)
+                prediction = action_classifier.infer(skeleton_seq)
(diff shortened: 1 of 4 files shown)


passalis (commented):
Thank you! Seems fine and fixes the issue mentioned in #259.

Can you also confirm that the code correctly handles cases where we do not find any skeleton in the first frame?

It seems that in  https://github.com/opendr-eu/opendr/blob/master/projects/perception/skeleton_based_action_recognition/demos/demo.py#L205 we need to perhaps include l. 208-211 inside `if counter > 0:`. Otherwise, ` skeleton_seq` will be undefined.

passalis:
> It seems that in https://github.com/opendr-eu/opendr/blob/master/projects/perception/skeleton_based_action_recognition/demos/demo.py#L205 we need to perhaps include l. 208-211 inside `if counter > 0:`. Otherwise, ` skeleton_seq` will be undefined.

I have fixed this. Please check it and feel free to revert if this is not correct.

ad-daniel (approved):
Thank you. I've taken the opportunity to fix cppcheck since it was recently upgraded on macos

passalis (approved):
Thank you!

PR #265, opened 2022-06-15
Update instructions and adapt nightly package tester

ad-daniel:
Ditto.


ad-daniel:
The CI appears to struggle to install the toolkit (full toolkit at once), it enters an endless loop of trying to solve dependencies.
I've tried on a clean ubuntu 20 docker image and the toolkit installed without an issue whatsoever, does anybody face the same issue as the CI?

passalis:
Hi Daniel, are you referring to pip installation? I was testing it yesterday in various systems using pip and I also noticed some warning messages and a increased installation time on some systems. I was going to investigate this.

passalis:
Are you referring to this https://github.com/opendr-eu/opendr/runs/6912594887?check_suite_focus=true ? For the specific one I suspect that there was some connectivity issues.

ad-daniel:
No, it was in the previous commits, but you're right, I might have fixed it unintentionally already by upgrading pip. I've re-run it, now to see if it shows up again. Basically it was pooch (from librosa) and llvmlite, but it might no longer be an issue

ad-daniel:
You can reproduce it right now or it happened before? Because I personally can't reproduce it locally

passalis:
No, I didn't manage to reproduce it locally. I tried several environments, skipping cached packages, etc, but it seems to install correctly. On colab.research.google.com I get an error where pip tries to download all available versions of a package. However, I am not 100% sure that this is indeed related or if it is another dependency that cannot be resolved.

passalis:
Did this got resolved by itself or I am missing something?

ad-daniel:
To an extent I think so, or at least it doesn't try to solve dependencies endlessly. It however seems to install torch 1.12 rather than 1.9 and for which `enable_onnx_checker` is no longer supported (since 1.10 at least), so those tests that include onnx fail. We can probably address that, but I don't understand from the logs why (or when) it installs torch 1.12 to begin. We have a hard constraint on 1.9 almost everywhere, and locally it installs that version

passalis:
I think I have managed to reproduce the error and probably the cause is indeed the version of torchvision/torchtext/torch audio:
```
torchvision 0.13.1+cu113 requires torch==1.12.1, but you have torch 1.9.0 which is incompatible.
torchtext 0.13.1 requires torch==1.12.1, but you have torch 1.9.0 which is incompatible.
torchaudio 0.12.1+cu113 requires torch==1.12.1, but you have torch 1.9.0 which is incompatible.
```

I am looking into this.

stefaniapedrazzi:
This PR is outdated and many of these changes have already been merged in the master branch recently.
I think we can close it.

PR #268, opened 2022-06-24
Fix EfficientPS

vniclas:
@ad-daniel I just realized that the latest merge of `develop` into `master` did not update the submodule. Strangely, in the `develop` branch it does correctly point to the latest version.


ad-daniel:
Dang, nice catch. We should probably trigger a v1.1.1 for this and the face_recognition compilation issue, @passalis what do you think?
If I recall correctly, we couldn't add the tests for this tool because the CI machines don't have gpu and some of the dependencies don't play nice without it. Do you think it would be possible to still have a test of some sort to validate the tool is working as intended? (even if it entails hard-coding/skipping the part that doesn't work without gpu)

vniclas:
Yes, that's correct. We can't run the tool in CI. The problem is that it already fails during the installation step. I'll think about whether we can add some other pseudo-tests.

ad-daniel (approved):
That would indeed be helpful, thank you!

passalis (approved):
Thank you! @ad-daniel  I agree, this one along the face recognition should be included in v1.1.1.

PR #269, opened 2022-06-24
Panoptic segmentation: download pretrained weights in ROS node

vniclas:
Question to reviewers: Would it be okay to store the downloaded model weights somewhere to avoid re-downloading upon each start of the node? Now, they are removed when the node is stopped.

Changelog:
- add an option to download pre-trained models (KITTI, Cityscapes) when starting the node
- change the order of required parameters matching other nodes

### DIFF
--- a/projects/opendr_ws/src/perception/scripts/panoptic_segmentation_efficient_ps.py
+++ b/projects/opendr_ws/src/perception/scripts/panoptic_segmentation_efficient_ps.py
@@ -154,7 +154,7 @@ class EfficientPsNode:
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
-    parser.add_argument('input_rgb_image_topic', type=str,
+    parser.add_argument('input_rgb_image_topic', type=str, default='/usb_cam/image_raw',
                         help='listen to RGB images on this topic')
     parser.add_argument('--checkpoint', type=str, default='cityscapes',
                         help='download pretrained models [cityscapes, kitti] or load from the provided path')
--- a/projects/opendr_ws/src/perception/scripts/panoptic_segmentation_efficient_ps.py
+++ b/projects/opendr_ws/src/perception/scripts/panoptic_segmentation_efficient_ps.py
@@ -97,7 +97,7 @@ class EfficientPsNode:
         """
         Subscribe to all relevant topics.
         """
-        rospy.Subscriber(self.input_rgb_image_topic, ROS_Image, self.callback)
+        rospy.Subscriber(self.input_rgb_image_topic, ROS_Image, self.callback, queue_size=1, buff_size=10000000)
 
     def _init_publisher(self):
         """
--- a/projects/opendr_ws/src/perception/scripts/panoptic_segmentation_efficient_ps.py
+++ b/projects/opendr_ws/src/perception/scripts/panoptic_segmentation_efficient_ps.py
@@ -66,7 +66,7 @@ class EfficientPsNode:
         self._learner = EfficientPsLearner(str(config_file))
 
         # Other
-        self._tmp_folder = Path(__file__).parent / 'tmp' / 'efficientps'
+        self._tmp_folder = Path(__file__).parent.parent / 'tmp' / 'efficientps'
         self._tmp_folder.mkdir(exist_ok=True, parents=True)
 
     def _init_learner(self) -> bool:
--- a/projects/opendr_ws/src/perception/scripts/panoptic_segmentation_efficient_ps.py
+++ b/projects/opendr_ws/src/perception/scripts/panoptic_segmentation_efficient_ps.py
@@ -67,15 +67,8 @@ class EfficientPsNode:
         self._learner = EfficientPsLearner(str(config_file))
 
         # Other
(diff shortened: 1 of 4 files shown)


passalis:
> Would it be okay to store the downloaded model weights somewhere to avoid re-downloading upon each start of the node? Now, they are removed when the node is stopped.

Yes, this is fine. Actually, this should be the expected behavior.

projects/opendr_ws/src/perception/scripts/panoptic_segmentation_efficient_ps.py
```diff
@@ -134,9 +152,11 @@ def callback(self, data: ROS_Image):
...
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
-    parser.add_argument('checkpoint', type=str, help='load the model weights from the provided path')
     parser.add_argument('image_topic', type=str, help='listen to images on this topic')
-    parser.add_argument('--heatmap_topic', type=str, help='publish the semantic and instance maps on this topic')
+    parser.add_argument('checkpoint', type=str,
+                        help='download pretrained models [cityscapes, kitti] or load from the provided path')
+    parser.add_argument('--heatmap_topic', type=str,
+                        help='publish the semantic and instance maps on this topic')
     parser.add_argument('--visualization_topic', type=str,
```
tsampazk, line 160: This could act similar to other topics and not publish anything if the topic is `None` and not use a separate flag.

Note that there is an issue with all other nodes i fixed, where the user is not able to provide `None` values through argparse, but i intend to fix this later, and will probably `sweep-fix` it across all toolkit ROS1/2 nodes as they should be technically the same.

tsampazk (commented):
I've added three comments regarding the node's interface and interaction with the user (class and argparse argument naming, default values, etc.). It would be helpful to refer to semantic segmentation bisenet node [here.](https://github.com/opendr-eu/opendr/blob/187813fee22c0c0f40d6edefbf3f3cb3d5721aec/projects/opendr_ws/src/perception/scripts/semantic_segmentation_bisenet.py)

As a note, you might need to add `, queue_size=1, buff_size=10000000)` arguments to `rospy.Subscriber` in `_init_subscribers` method, and test whether there is a delay between input and output, greater than the time it takes to process each frame and adjust the buffer size accordingly, relating to #275. In all other nodes i've worked on there's an obvious big delay when the node was running slowly (e.g. on cpu). You might be able to test it by adding `sleep` in the callback to artificially make it slow. You can find a more thorough explanation in related issue.

projects/opendr_ws/src/perception/scripts/panoptic_segmentation_efficient_ps.py
tsampazk: I dug deeper in the code and found out this is indeed a flag for the learner's visualize method and not what i thought before. Visualization topic seems to work as expected. Nevertheless, do you think we can remove it altogether and pass a True value to the visualization method so we get the maximum information out of it and simplify the interface?

As a side note, visualization seems to not work with different sizes of input images, can you confirm this? I try to run the node and provide images  through a 640x480 webcam and i get a `ValueError: images do not match` error on `...efficient_ps/efficient_ps_learner.py", line 547, in visualize panoptics_img = PilImage.blend(`...`.

tsampazk:
Hey @vniclas, just a heads up, i edited the links in the first comment to point to the updated bisenet node. It should now be more similar in terms of naming and output to the panoptic node and i think it makes it easier to make them more consistent.

vniclas:
Hi @tsampazk, thanks for your in-depth review. I addressed all the comments except for having a default value for the input RGB image topic. There, I disagree with you. In my opinion, having a default value for this is more confusing than forcing the user to state the topic name explicitly.

Thanks for pointing out the issue with images of different sizes. That should be fixed now.

tsampazk:
Thank you @vniclas for the changes and fixes! 

> There, I disagree with you. In my opinion, having a default value for this is more confusing than forcing the user to state the topic name explicitly.

I can see your point and you are right that it might cause confusion when running the nodes in different environments other than using a default webcam. Maybe we should discuss it in the next technical meeting and ask the use-case partners what they prefer, so i can change it in all the others nodes for consistency.

As a side note, did you manage to test the delay caused by the queues, because i can see that the [subscriber](https://github.com/opendr-eu/opendr/blob/fix/ros-efficientps/projects/opendr_ws/src/perception/scripts/panoptic_segmentation_efficient_ps.py#L100) still uses the default queue size, maybe you missed my comment [here](https://github.com/opendr-eu/opendr/pull/269#pullrequestreview-1080247409)?

Other than that, i think the PR is good to go after we settle the input argument thing.

vniclas:
I did miss that specific comment. I will look into it as soon as possible and then get back to you. Thanks!

vniclas:
I set the queue size to 1 and increased the buffer size as proposed by you. I measured the output frequency and the run time of the callback (convert ROS message to image, network forward pass, visualization). They correspond to each other, i.e., there is no notable delay.

vniclas:
I added the default value `/usb_cam/image_raw` for the input RGB image stream and updated the documentation accordingly.

I believe, this PR is ready for the final review now.

tsampazk (approved):
Thank you @vniclas!

passalis (approved):
Thank you!

PR #279, opened 2022-07-13
Add ROS2 face recognition and fall detection nodes

tsampazk:
This PR adds the Face Recognition and Fall Detection ROS2 nodes, with appropriate bridge methods as well as setup.py changes.

It also fixes minor problems with arguments in earlier ROS2 nodes (face detection retina face and object detection 2d ssd nodes), and reorders the to/from image methods in bridge.py, to improve internal consistency of the file (for any type, first comes the 'to' method, then the 'from' method).

- [x] Add output annotated image for FR node
- [x] Add fall detection node


passalis (approved):
This PR seems fine to me. I think we can ignore the failing test, since this PR is not the issue for this (we have the same failure also in the master and in other PRs as well).

PR #284, opened 2022-08-02
Added Frame ID for visulaization purposes

GehadAboHarga:
# Description

Point cloud ROS node published data without frame id so it was not possible to visualize it on rviz. so I added frame id to the data.

## Type of change

-  [x] Quick Fix

### DIFF
--- a/projects/opendr_ws/src/perception/scripts/point_cloud_dataset.py
+++ b/projects/opendr_ws/src/perception/scripts/point_cloud_dataset.py
@@ -53,6 +53,8 @@ class PointCloudDatasetNode:
             message = self.bridge.to_ros_point_cloud(
                 point_cloud
             )
+            message.header.frame_id = "velodyne"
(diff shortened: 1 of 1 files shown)


projects/opendr_ws/src/perception/scripts/point_cloud_dataset.py
```diff
@@ -53,6 +53,8 @@ def start(self):
             message = self.bridge.to_ros_point_cloud(
                 point_cloud
             )
+            message.header.frame_id = "velodyne"
```
passalis, line 56: Thank you for your contribution. We can rename the `frame_id` to something more generic.

```suggestion
            message.header.frame_id = "lidar_dataset"
```

passalis:
Since there was no update and this is a low priority update/feature, I am closing this PR until deciding how/if this should be handled.

PR #286, opened 2022-08-18
Ros nodes e2e planner

halil93ibrahim:
Hi,

I have started implementing both ROS1 and ROS2 nodes for end to end planner under this branch. I have a question regarding the ROS2 node implementation.

The gym environment I have provided with the learner is currently communicates with Webots and mavros through ROS1. So, the ROS nodes are supposed to interact with this environment. What should be the right strategy when I implement the ROS2 node. Should I have optional ROS2 communication for the gym environment? Or, should I bridge the communication between ROS1 and ROS2?


passalis:
Hi Hallil, if you can directly support ROS2 (instead of using a bridge) without much effort then I think this would be better. @ad-daniel @stefaniapedrazzi what do you think?

I also converted this PR to draft until it is ready for review. If this is not the case, please revert it.

halil93ibrahim:
Hi, I will try to put everything in ROS2.

halil93ibrahim:
Hi,

I have updated the end-to-end planner tool to be included the work in my recent [paper](https://www.mdpi.com/2218-6581/11/5/109). I have also completed ROS1 node which publishes the output of the planner as a target waypoint,

Unfortunately, I could not complete ROS2 node, because of the issues faced with 2021a version of Webots and ROS2. While trying to solve this, we could merge the already implemented part and complete ROS2 part in another pr. What do you think? In that case, should I target another branch or, is this PR works?

tsampazk:
> What do you think? In that case, should I target another branch or, is this PR works?

Hello @halil93ibrahim! I think that it would be better to target develop with this PR as it will no longer include a ROS2 node and it involves changes in a lot of files other than the ROS2 workspace. When you manage to make the ROS2 node work, you can add it later in a new PR that will target the ros2 branch.

halil93ibrahim:
Ok, I am closing this PR and will create a new one targeting develop.

tsampazk:
Oh you can retarget the PR by editing it, but it's ok if you already done the work.

halil93ibrahim:
I did not know and had already created a new PR.

tsampazk:
> I did not know and had already created a new PR.

It's ok no problem, sorry for not mentioning it originally!

PR #287, opened 2022-08-22
Fix detectron2

passalis:
This PR updated the toolkit to points to a specific (and tested) commit of detectron2 in order to avoid breaking the installation when changes occur in detectron repository.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -48,7 +48,7 @@ if [[ "${OPENDR_DEVICE}" == "gpu" ]]; then
   echo "[INFO] Replacing torch==1.9.0+cu111 to enable CUDA acceleration."
   pip3 install torch==1.9.0+cu111 torchvision==0.10.0+cu111 torchaudio==0.9.0 -f https://download.pytorch.org/whl/torch_stable.html
   echo "[INFO] Reinstalling detectronv2."
-  pip3 install 'git+https://github.com/facebookresearch/detectron2.git'
+  pip3 install 'git+https://github.com/facebookresearch/detectron2.git@5aeb252b194b93dc2879b4ac34bc51a31b5aee13'
 fi
 
 make libopendr
--- a/docs/reference/installation.md
+++ b/docs/reference/installation.md
@@ -103,6 +103,10 @@ pip install mxnet-cu112==1.8.0post0
 pip install opendr-toolkit-engine
 pip install opendr-toolkit
 ```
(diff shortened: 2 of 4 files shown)


passalis:
The single demonstration grasp issue seems to have been resolved (it was caused by an updated version of detectron). However, there seems to be also one other issue with a simulation tool. I am also looking into this. I am converting this draft until this is also sorted out.

charsyme:
@passalis https://github.com/opendr-eu/opendr/blob/3a9a119f2f4391c3d9ba5f2f998400ffb61f72cf/src/opendr/simulation/human_model_generation/utilities/PIFu/lib/mesh_util.py#L45 In newer scikit-image versions, this function is deprecated and needs to change to  "verts, faces, normals, values = measure.marching_cubes(sdf, 0.5)"

passalis:
Thank you @charsyme. I think this is now ready. I will wait for the tests to pass and I will mark this ready for review.

passalis:
This is now ready for review.

passalis:
I think the failures are unrelated to this PR. I will try running the tests once more.

passalis:
I think we are in a deadlock... This PR needs #207 for passing 3D detection tests (which now seem to fail very often), while #207 needs this PR for solving grasping and simulation failures. 2D tracking test failure is due to a download issue. I will try running the tests once again...

stefaniapedrazzi:
> I think we are in a deadlock... This PR needs #207 for passing 3D detection tests (which now seem to fail very often), while #207 needs this PR for solving grasping and simulation failures. 2D tracking test failure is due to a download issue. I will try running the tests once again...

Given that the master branch is currently in add state anyway, personally I think it is ok to merge this PR and then work at the remaining errors in a different PR.

stefaniapedrazzi (approved):
Thank you!

ad-daniel:
Should we perhaps disable fast-failure in the CI? Currently as soon as one test fails, the entire job is canceled.
This is positive as it frees the machines for other PRs, but at the same when we have issues with a different test (for instance out of memory), you can't even know if your change works since others might fail before yours gets the chance to run

ad-daniel (approved):
It doesn't appear there are side-effects from the change to the requirements. All perception/object_detection_2d tests failed (were canceled after 6h) but as you mentioned that's another issue. The detectron fix seems to work as intended. Thanks!

stefaniapedrazzi (approved):

Issue #275, opened 2022-07-04
ROS nodes delay


tsampazk:
### General Description
Some ROS nodes exhibit delay between the current frame and the frame that is being published.

### Cause
This is because in most (all?) nodes the subscriber which uses the callback that performs inference, has a `queue_size` of `10`. If the callback inference is slow, the subscriber keeps the incoming frames and processes them all one-by-one. When processing of one frame ends, the callback grabs an old frame from inside the queue and so on, causing the delay in the output.

### Expected behaviour and solution
The expected behaviour would be to process the newest, most update-to-date frame every time the callback grabs a frame for inference and not use older frames. This is done by setting the `queue_size` to `1`, but to properly work in ROS 1 you also need to set the `buffer_size` accordingly. I found that a value of `5000000` for the buffer size works well on object detection 2D. The delay is minimized to whatever time it takes for one frame to process which is expected.

Edit: ~~Some~~ All nodes require a larger buffer size to minimize delay, e.g. `10000000`.

Edit: **For ROS2 it is not required to set the buffer size, just add `1` as the queue size as seen [here](https://github.com/opendr-eu/opendr/blob/b9dab8c0764143d9817944851defa8502750a146/projects/opendr_ws_2/src/opendr_perception/opendr_perception/face_detection_retinaface_node.py#L53-L63).**

tsampazk:
I'll document my understanding below for future reference. After the fixes you should expect a delay of 2 frames. This is because, as far as i can tell, **we have to work with a minimum buffer of size `1`**, plus the frame that was already being processed. In more detail:

![ROS_delay_diagram](https://user-images.githubusercontent.com/27914645/185743734-8f996c76-a376-4cd0-a703-6367a48f34f9.jpg)

At times `t`, `t + 1`, etc. processed frames are being published. Keep in mind that frames are coming in as fast as the camera being used allows. Irrelevant frames are omitted.

1. A **reference visible movement** starts at `tm` and is completed before `t + 1`, e.g. a subject person raised their hand quickly (movement is complete before `t + 1`) and holds it in position until `t + 1`. 
2. The critical frame `Fm` that shows the person with their hand raised is captured at `t+1` and is stored in the buffer (that's when `F2` is taken from the buffer to process and the buffer is empty). At the same time `F1` is published. 
3. `F2` is published at `t+2` and processing of `Fm` begins. 
4. Finally, `Fm` is published at `t+3` **with a total delay of the time it takes for 2 frames to process.**

[...]

PR #295, opened 2022-09-02
Ros2 gem

jelledouwe:
This PR contains the ros2 node for the multimodal object detector GEM. This PR also contains changes to the ROS1 gem node that address the feedback from PAL.


projects/opendr_ws/src/perception/scripts/object_detection_2d_gem.py
```diff
@@ -176,25 +221,25 @@ def callback(self, msg_rgb, msg_ir):
...
+            device = "cuda"
         else:
             print("GPU not found. Using CPU instead.")
-            device = 'cpu'
+            device = "cpu"
     except:
-        device = 'cpu'
+        device = "cpu"
     detection_estimation_node = GemNode(device=device)
     detection_estimation_node.listen()
```
tsampazk, line 245: Main should have the same argparse logic as in the ROS2 node, so as to enable users to modify the behaviour of the node.

projects/opendr_ws/src/perception/scripts/object_detection_2d_gem.py
```diff
@@ -121,11 +174,12 @@ def __init__(self,
...
-        self.gem_learner.fusion_method = 'sc_avg'
+        self.gem_learner = GemLearner(
+            backbone=model_backbone,
+            num_classes=7,
+            device=device,
+        )
+        self.gem_learner.fusion_method = "sc_avg"
         self.gem_learner.download(path=".", verbose=True)
 
         # Subscribers
```
tsampazk, line 185: (Can't add comment directly to unchanged lines)

Both subscribers should have `queue_size=1, buff_size=10000000` arguments added to avoid delays, relevant issue #275

projects/opendr_ws_2/src/opendr_perception/opendr_perception/object_detection_2d_gem_node.py
```diff
@@ -0,0 +1,281 @@
...
+            backbone=model_backbone,
+            num_classes=7,
+            device=device,
+        )
+        self.gem_learner.fusion_method = "sc_avg"
+        self.gem_learner.download(path=".", verbose=True)
+
+        # Subscribers
+        msg_rgb = message_filters.Subscriber(self, ROS_Image, input_rgb_image_topic)
+        msg_ir = message_filters.Subscriber(self, ROS_Image, input_infra_image_topic)
```
tsampazk, line 193: Similar to ROS1 node, both subscribers should have their queues set to 1 to avoid delays (issue #275).
```suggestion
        msg_rgb = message_filters.Subscriber(self, ROS_Image, input_rgb_image_topic, 1)
        msg_ir = message_filters.Subscriber(self, ROS_Image, input_infra_image_topic, 1)
```

tsampazk (changes requested):
Hey @jelledouwe, thank you for the changes and additions! ROS2 node is mostly good to go apart from some minor comments. Some changes are required in the ROS1 node so it is more consistent with the rest of the object detection nodes as well as the gem ROS2 node.

Note that both nodes require some changes at the queue sizes of the subscribers to avoid delays, relating to issue #275 .

projects/opendr_ws/src/perception/scripts/object_detection_2d_gem.py
tsampazk: Subscribers still don't have their queues set. Have you tested whether the node has delay issues? If not, then the queues are not needed.

projects/opendr_ws_2/src/opendr_perception/opendr_perception/object_detection_2d_gem_node.py
```diff
@@ -0,0 +1,293 @@
...
+            backbone=model_backbone,
+            num_classes=7,
+            device=device,
+        )
+        self.gem_learner.fusion_method = "sc_avg"
+        self.gem_learner.download(path=".", verbose=True)
+
+        # Subscribers
+        msg_color = message_filters.Subscriber(self, ROS_Image, input_color_image_topic)
+        msg_ir = message_filters.Subscriber(self, ROS_Image, input_infra_image_topic)
```
tsampazk, line 191: Same as the ROS1 node, the queues are still not set.

projects/opendr_ws_2/src/opendr_perception/opendr_perception/object_detection_2d_gem_node.py
```diff
@@ -0,0 +1,293 @@
...
+from opendr.perception.object_detection_2d import draw_bounding_boxes
+
+
+class ObjectDetectionGemNode(Node):
+    def __init__(
+        self,
+        input_color_image_topic,
+        input_infra_image_topic,
+        output_color_image_topic,
+        output_infra_image_topic,
```
tsampazk, line 39: In an earlier commit "color" was changed to "rgb" as far as i can tell, but the changes were reverted, same as in the ROS1 node.

projects/opendr_ws/src/perception/scripts/object_detection_2d_gem.py
```diff
@@ -18,43 +18,43 @@
...
-                 device="cuda",
-                 pts_color=None,
-                 pts_infra=None,
-                 ):
+class ObjectDetectionGemNode:
+    def __init__(
+        self,
+        input_color_image_topic="/camera/color/image_raw",
+        input_infra_image_topic="/camera/infra/image_raw",
+        output_color_image_topic="/opendr/color_detection_annotated",
```
tsampazk, line 36: In an earlier commit "color" was changed to "rgb" as far as i can tell, but the changes were reverted.

projects/opendr_ws/src/perception/scripts/object_detection_2d_gem.py
```diff
@@ -143,30 +199,22 @@ def listen(self):
...
         """
         Callback that process the input data and publishes to the corresponding topics
-        :param msg_rgb: input color image message
-        :type msg_rgb: sensor_msgs.msg.Image
+        :param msg_color: input color image message
+        :type msg_color: sensor_msgs.msg.Image
         :param msg_ir: input infrared image message
         :type msg_ir: sensor_msgs.msg.Image
         """
+        rospy.logwarn("Entered callback!")
```
tsampazk, line 210: This logwarn looks like it is needed for debugging purposes. If that's the case please remove it as it will produce lot's of unnecessary logging for the end-user.

projects/opendr_ws/src/perception/scripts/object_detection_2d_gem.py
```diff
@@ -175,26 +223,66 @@ def callback(self, msg_rgb, msg_ir):
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--input_color_image_topic", help="Topic name for input color image", type=str, default="/camera/color/image_raw"
+    )
+    parser.add_argument(
+        "--output_color_image_topic",
+        help="Topic name for output annotated color image",
+        type=str,
+        default="/opendr/color_objects_annotated",
+    )
```
tsampazk, line 246: Similar to the constructor arguments, "color" should be "rgb" for consistency with other nodes. 

Moreover, other object detection nodes that only output annotated RGB images use the  "image_objects_annotated" default value for the output topic. This node has both infrared and RGB annotated images, so for consistency i would suggest that it could become "rgb_image_objects_annotated" and similarly, the infrared output topic should be "infra_image_objects_annotated". Whatever default value you choose should be also changed in the ctor arguments, docstrings, etc.

projects/opendr_ws_2/src/opendr_perception/opendr_perception/object_detection_2d_gem_node.py
```diff
@@ -0,0 +1,293 @@
...
+    )
+    parser.add_argument(
+        "--input_infra_image_topic", help="Topic name for input infra image", type=str, default="/camera/infra/image_rect_raw"
+    )
+    parser.add_argument(
+        "--output_infra_image_topic",
+        help="Topic name for output annotated infra image",
+        type=str,
+        default="/opendr/infra_objects_annotated",
+    )
```
tsampazk, line 250: ROS1 node comment for "color"->"rgb" and default value for topics also applies here.

tsampazk (changes requested):
Thanks @jelledouwe for the fixes so far! There have been some suggested changes that were applied and later reverted. I also added some more comments on argument naming and default values that mainly have to do with consistency across the object detection nodes.

Please also take a look at the unresolved comments from the previous review.

jelledouwe:
> i noticed is that the default topic value fo

Hi, thanks for the review. That is well spot! The reason why I did this, is that the default values now work well with RealSense cameras which are probably the most popular rgb+depth cameras in robotics. Since the topic name is changed in the ROS 2 RealSense camera package, I also changed it in the OpenDR ROS2 node. I would suggest to keep it this way, such that the node works well with RealSense cameras.

tsampazk (approved):
> Since the topic name is changed in the ROS 2 RealSense camera package, I also changed it in the OpenDR ROS2 node. I would suggest to keep it this way, such that the node works well with RealSense cameras.

Oh ok, sorry i was not aware, it makes sense then, we can keep it this way! Thanks.

PR #306, opened 2022-09-14
ROS1 nodes - None argument fix

tsampazk:
Introduced a lambda function in the `type` argument of argparse to handle the passing of `None` value, along some other fixes in fall detection node and some minor optimizations.

Previously, users were unable to disable an output topic by passing `None` through the console when running the nodes, as it was passed `str` instead of `NoneType`.


projects/opendr_ws/src/perception/scripts/fall_detection.py
```diff
@@ -106,23 +106,28 @@ def callback(self, data):
         image = image.opencv()
 
         bboxes = BoundingBoxList([])
+        fallen_pose_id = 0
         for detection in detections:
             fallen = detection[0].data
             pose = detection[2]
+            x, y, w, h = get_bbox(pose)
```
stefaniapedrazzi, line 113: All the changes seems very good.

Just a minor question: why do you need to compute `pose` and the bbox outside the `if fallen == 1` block?
They don't seem to be used in other places.

tsampazk: You're right, i think it used to make sense in earlier versions! I moved them both inside the if block. `detection[0].data` and `detection[2]` could be used straight away also instead of assigning them to `fallen and `pose`, but i chose to keep them as they are to make the code more self-explanatory.

stefaniapedrazzi (approved):
Thank you!

passalis (approved):
Thank you!

PR #311, opened 2022-09-19
Fix cppcheck 2.9 errors

stefaniapedrazzi:
Fix errors in `test_sources` test due to upgrade to cppcheck 2.9 on macOS.

In the CI the latest cppcheck version available on the system is used.
Given that the changes required to be compliant with cppcheck 2.9 just a couple I would say that it is better to continue with the latest version and fix the sources.


passalis (approved):
Thank you!

ad-daniel (approved):
Thank you

PR #313, opened 2022-09-22
modify TypeError

jjaegii:
[ TypeError: only integer scalar arrays can be converted to a scalar index ] at 'selected_poses.append(poses[index])' in 'select_2_poses'

### DIFF
--- a/projects/perception/skeleton_based_action_recognition/demos/demo.py
+++ b/projects/perception/skeleton_based_action_recognition/demos/demo.py
@@ -35,7 +35,7 @@ class VideoReader(object):
     def __init__(self, file_name):
         self.file_name = file_name
         try:  # OpenCV needs int to read from webcam
-            self.file_name = int(file_name)
+            self.file_name = file_name
         except ValueError:
             pass
 
@@ -61,7 +61,6 @@ def tile(a, dim, n_tile):
     order_index = torch.LongTensor(np.concatenate([init_dim * np.arange(n_tile) + i for i in range(init_dim)]))
     tiled_a = torch.index_select(a, dim, order_index)
     return tiled_a.numpy()
-
 
 def pose2numpy(args, num_current_frames, poses_list):
     C = 2
@@ -92,6 +91,7 @@ def pose2numpy(args, num_current_frames, poses_list):
     return skeleton_seq
 
 
+# TypeError: only integer scalar arrays can be converted to a scalar index (complete)
 def select_2_poses(poses):
     selected_poses = []
     energy = []
@@ -100,7 +100,8 @@ def select_2_poses(poses):
         energy.append(s)
     energy = np.array(energy)
     index = energy.argsort()[::-1][0:2]
-    selected_poses.append(poses[index])
+    # selected_poses.append(poses[index])
+    selected_poses.append(np.array(poses)[index])
     return selected_poses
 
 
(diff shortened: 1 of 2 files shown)


projects/perception/skeleton_based_action_recognition/demos/demo.py
```diff
@@ -92,6 +91,7 @@ def pose2numpy(args, num_current_frames, poses_list):
     return skeleton_seq
 
 
+# TypeError: only integer scalar arrays can be converted to a scalar index (complete)
```
stefaniapedrazzi, line 94: This comment seems more a development message and should probably be removed.

projects/perception/skeleton_based_action_recognition/demos/demo.py
```diff
@@ -100,7 +100,8 @@ def select_2_poses(poses):
         energy.append(s)
     energy = np.array(energy)
     index = energy.argsort()[::-1][0:2]
-    selected_poses.append(poses[index])
+    # selected_poses.append(poses[index])
```
stefaniapedrazzi, line 103: This line should be removed

stefaniapedrazzi (changes requested):
Thank you. The fix is fine.
But there are some additional changes that doesn't seem needed.

projects/perception/skeleton_based_action_recognition/demos/demo.py
```diff
@@ -35,7 +35,7 @@ class VideoReader(object):
     def __init__(self, file_name):
         self.file_name = file_name
         try:  # OpenCV needs int to read from webcam
-            self.file_name = int(file_name)
+            self.file_name = file_name
```
passalis, line 38: This try block will always succeed after the change. So, if we do not need this it should be altogether removed. Why is this change needed?

jjaegii: because you can bring not only webcams but also other videos. for example, rtsp(cctv) streams.

projects/perception/skeleton_based_action_recognition/demos/demo.py
jjaegii: i will fix it soon, thank you

jjaegii:
change cuda to cpu,
remove useless comment.

projects/perception/skeleton_based_action_recognition/demos/demo.py
passalis: In the modified version of the code the `try` block can be removed altogether, since the assignment has already been done in l.36. Removing lines 37-40 in the current version (without the int cast) will not change anything.

However, I was trying to understand why this block exists in the first place.   To the best of my understanding, the behavior of VideoCapture changes according to the data types it receives (https://docs.opencv.org/3.4/d8/dfe/classcv_1_1VideoCapture.html). When an `int` is passed, then it tries to open the corresponding camera. On the other hand, when a string is passed, then it tries to open the corresponding stream or file. The `try`... `except` block is essentially trying to cope this behavior - in a not very clear way (especially given that the argument is called `file_name`). If the supplied `file_name` string can be cast into an `int`, then it implicitly forces opencv to read from the corresponding webcam. Of course, if there is a file with no suffix and can be cast into a number, then this code will not read from the file and it will try to open the corresponding webcam. However, in any other case it should function correctly. For example, you shouldn't have issues with reading from rtsp streams using the existing version of the code.

jjaegii: oh, i see!

Issue #314, opened 2022-09-22
Missing "panopticapi" module on single tool install using pip


stefaniapedrazzi:
When installing the single `opendr-toolkit-object_detection_2d` following the [instructions](https://github.com/opendr-eu/opendr/blob/master/docs/reference/installation.md#installing-only-a-particular-tool-using-pip-cpugpu) I get this error on import:
```
Python 3.8.14 (default, Sep  8 2022, 00:02:07) 
[GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from opendr.perception.object_detection_2d import RetinaFaceLearner
/home/stefi/develop/opendr/venv/lib/python3.8/site-packages/gluoncv/__init__.py:40: UserWarning: Both `mxnet==1.8.0` and `torch==1.9.0+cu102` are installed. You might encounter increased GPU memory footprint if both framework are used at the same time.
  warnings.warn(f'Both `mxnet=={mx.__version__}` and `torch=={torch.__version__}` are installed. '
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/stefi/develop/opendr/venv/lib/python3.8/site-packages/opendr/perception/object_detection_2d/__init__.py", line 2, in <module>
    from opendr.perception.object_detection_2d.detr.detr_learner import DetrLearner
  File "/home/stefi/develop/opendr/venv/lib/python3.8/site-packages/opendr/perception/object_detection_2d/detr/detr_learner.py", line 31, in <module>
    from opendr.perception.object_detection_2d.detr.algorithm.util.detect import detect
  File "/home/stefi/develop/opendr/venv/lib/python3.8/site-packages/opendr/perception/object_detection_2d/detr/algorithm/util/detect.py", line 24, in <module>
    from panopticapi.utils import rgb2id
ModuleNotFoundError: No module named 'panopticapi'
```

I'm further investigating if I did something wrong during the installation.

stefaniapedrazzi:
To make it work I have to install manually:
* tensorboardX
* git+https://github.com/cocodataset/panopticapi.git#egg=panopticapi

PR #321, opened 2022-09-28
Fix missing Python dependency in opendr-toolkit-object_detection_2d

stefaniapedrazzi:
Address  #314:
When installing the pip package `opendr-toolkit-object_detection_2d` Python complaints that `tensorboardX` is not installed and it is not possible to use the module.

`tensorboardX` is used here:
https://github.com/opendr-eu/opendr/blob/56d398799f52cc60f8ce6efed19854c3485c38bb/src/opendr/perception/object_detection_2d/nms/seq2seq_nms/seq2seq_nms_learner.py#L32


ad-daniel (approved):
Thank you

passalis (approved):
Thank you!

PR #326, opened 2022-09-30
Update har onnx inference and benchmark scripts

LukasHedegaard:
This PR rectifies som issues with ONNX inference and updates benchmark scripts


tsampazk (commented):
Thanks for the updates @LukasHedegaard! In the two benchmark scripts, on the changed portions there are several commented out lines. Are they actually needed or could we remove them altogether?

LukasHedegaard:
They are nice to have for setting max batch sizes in case throughput is to be measured

tsampazk (approved):

ad-daniel (approved):
Yes I think we can leave them, thank you!

PR #337, opened 2022-10-15
ROS2 for heart anomaly detection

minhquoc0712:
Add:

- ROS2 implementation for heart anomaly detection.
- Methods to handle time series and categories in `ROS2Bridge`.

### DIFF
--- a/projects/opendr_ws_2/src/opendr_perception/opendr_perception/heart_anomaly_detection_node.py
+++ b/projects/opendr_ws_2/src/opendr_perception/opendr_perception/heart_anomaly_detection_node.py
@@ -1,0 +1,127 @@
+#!/usr/bin/env python
+# Copyright 2020-2022 OpenDR European Project
+#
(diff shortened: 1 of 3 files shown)


minhquoc0712:
I initialize `HeartAnomalyNode` under the name of `opendr_heart_anomaly_detection` to be consistent with the ROS node name of this algorithm. But I see that other nodes have a different naming convention such as `object_detection_centernet_node`. Do I need to change my node name?

tsampazk:
> I initialize `HeartAnomalyNode` under the name of `opendr_heart_anomaly_detection` to be consistent with the ROS node name of this algorithm. But I see that other nodes have a different naming convention such as `object_detection_centernet_node`. Do I need to change my node name?

You're right, in ROS2 the naming is a bit different. For now, please try to use names that are consistent with the rest of the ROS2 nodes.

tsampazk (approved):
Thank you, new node looks good!

PR #340, opened 2022-10-17
ROS2 speech command recognition

minhquoc0712:
Implementation details:

- Add ROS2 for speech command recognition.
- Change default audio topic from `/audio/audio` to `/audio`.
- Add `from_category_to_rosclass` in `ROS2Bridge`.

### DIFF
--- a/projects/opendr_ws_2/src/opendr_perception/opendr_perception/speech_command_recognition_node.py
+++ b/projects/opendr_ws_2/src/opendr_perception/opendr_perception/speech_command_recognition_node.py
@@ -146,4 +146,4 @@ def main(args=None):
 
 
 if __name__ == "__main__":
-	main()
+    main()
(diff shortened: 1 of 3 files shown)


minhquoc0712:
Do I need to include a clone version of this audio capture repo: https://github.com/ros-drivers/audio_common/tree/ros2.

tsampazk:
> Do I need to include a clone version of this audio capture repo: https://github.com/ros-drivers/audio_common/tree/ros2.

There's no need to include it in our repository, just use it locally to test your node. The idea is similar to how we use the usb cam packages. If you can, post some brief information of what package you used and how you used it here, so maybe i can include this info in the README.

minhquoc0712:
Note for installing the `audio_common` package in ROS2:

1. Clone the following [ROS2 branch of `audio_common` package](https://github.com/ros-drivers/audio_common/tree/ros2) to the `opendr_ws_2/src` directory.
2. Install necessary packages: `rosdep install -i --from-path src --rosdistro foxy -y` 
3. Run ROS2 node to capture audio: `ros2 run audio_capture audio_capture`. The output audio topic is `/audio`.

tsampazk (approved):
Thanks, node looks really good and runs smoothly with the suggested audio package. Just a question, do you know where i can find the id-class correspondence? Is it documented somewhere?

PR #345, opened 2022-10-20
landmark_based_facial_expression_recognition ROS2 node

negarhdr:
This branch adds ROS2 node for landmark_based_facial_expression_recognition tool.


tsampazk (changes requested):
Thanks for the new node, some minor issues, same as the HAR node. I also added the node in setup.py.

PR #348, opened 2022-11-09
Allow to not use a pretrained FairMOT backbone

iliiliiliili:
This PR fixes the dependency on the external model downloads by not requiring a pretrained backbone if you intend to test using the trained model from the FTP server.


src/opendr/perception/object_tracking_2d/fair_mot/algorithm/lib/models/networks/dlav0.py
```diff
@@ -392,14 +392,17 @@ def forward(self, x):
 
             return x
 
-    def load_pretrained_model(self, data="imagenet", name="dla34", hash="ba72cf86"):
+    def load_pretrained_model(self, load=True, data="imagenet", name="dla34", hash="ba72cf86", num_classes=None):
```
tsampazk, line 395: With the new changes, i think that this method might need to be renamed as it no longer just loads a pretrained model, but might just create a model based on its `load` argument which is provided with the value of `pretrained` from methods below.

Would it be better to name it `load_model` and the argument load renamed to `pretrained`?

src/opendr/perception/object_tracking_2d/fair_mot/algorithm/lib/models/networks/pose_dla_dcn.py
```diff
@@ -372,13 +372,16 @@ def forward(self, x):
...
         return y
 
-    def load_pretrained_model(self, data="imagenet", name="dla34", hash="ba72cf86"):
-        if name.endswith(".pth"):
-            model_weights = torch.load(data + name)
-        else:
-            model_url = get_model_url(data, name, hash)
-            model_weights = model_zoo.load_url(model_url)
-        num_classes = len(model_weights[list(model_weights.keys())[-1]])
+    def load_pretrained_model(self, load=True, data="imagenet", name="dla34", hash="ba72cf86", num_classes=1000):
```
tsampazk, line 375: The comment from dlav0.py also applies here.

tsampazk (commented):
Thanks you for these changes, i have a minor comment. Other than that looks good!

ad-daniel (approved):
Thank you, hopefully it will reduce the number of failures

passalis (approved):
Thank you!

PR #351, opened 2022-11-22
fix: eagerx version demos

jelledouwe:
The demos for eagerx broke, because the eagerx version wasn't fixed. This is solved by this PR.


jelledouwe:
> It failed to install for me due to: `configparser.DuplicateOptionError: While reading from '/home/nctrl/opendr/projects/control/eagerx/dependencies.ini' [line 3]: option 'python' in section 'runtime' already exists`
> 
> Works for me after the change. Thank you

Thanks for the change, that was indeed the intended update.

tsampazk (approved):
Thank you!

PR #354, opened 2022-11-23
LW OpenPose: Added torch.set_flush_denormal(True) for major increase in cpu performance

tsampazk:
Added `torch.set_flush_denormal(True)` when using cpu, increasing performance in terms of fps, matching those [reported on the doc](https://github.com/opendr-eu/opendr/blob/master/docs/reference/lightweight-open-pose.md#performance-evaluation). Tested on a couple of local machines, fps increased from ~0.4 to ~8 fps and from ~1 to ~13 fps, when using the demo benchmarking script. Similar results on webcam demo.

Renamed some variables to follow snake_case naming instead of mixedCase.


ad-daniel (approved):
Thank you

passalis (approved):
Thank you!

PR #358, opened 2022-11-28
ROS2 node for e2e planner and environment upgrade to Webots R2022b

halil93ibrahim:
Hello,
This PR implements the ROS2 node for end-to-end planner, similar to ROS1 node. It also updates the Webots environment to the version of R2022b, which was previously R2021a. There are small fixes in the code to adapt to the version change. ros-numpy package dependency is also replaced since it cannot run along with ROS2 scripts.
A Webots launcher is also provided along with the ROS2 node, related files are also inside opendr_planning package since this world is specifically for the end-to-end planning node. Do you think this is ok?


src/opendr/planning/end_to_end_planning/__init__.py
```diff
@@ -1,4 +1,6 @@
 from opendr.planning.end_to_end_planning.e2e_planning_learner import EndToEndPlanningRLLearner
-from opendr.planning.end_to_end_planning.envs.UAV_depth_planning_env import UAVDepthPlanningEnv
+import os
+if os.environ.get("ROS_DISTRO") == "noetic":
```
ad-daniel, line 3: we should support melodic too, I don't think there's anything about what you do which would be different so it should work out of the box

ad-daniel (changes requested):
Thank you, I have a few comments

projects/opendr_ws_2/src/opendr_planning/launch/e2e_planning_robot_launch.py
```diff
@@ -0,0 +1,55 @@
...
+# limitations under the License.
+
+import os
+import pathlib
+import launch
+from launch_ros.actions import Node
+from launch import LaunchDescription
+from ament_index_python.packages import get_package_share_directory
+from webots_ros2_driver.webots_launcher import WebotsLauncher, Ros2SupervisorLauncher
+
```
ad-daniel, line 23:
```suggestion
from webots_ros2_driver.utils import controller_url_prefix

```

ad-daniel: it's a recent change, but this will make it work under more platforms other than linux

projects/opendr_ws_2/src/opendr_planning/launch/e2e_planning_robot_launch.py
```diff
@@ -0,0 +1,55 @@
...
+        world=os.path.join(package_dir, 'worlds', 'train-no-dynamic-random-obstacles.wbt')
+    )
+
+    ros2_supervisor = Ros2SupervisorLauncher()
+
+    my_robot_driver = Node(
+        package='webots_ros2_driver',
+        executable='driver',
+        output='screen',
+        additional_env={'WEBOTS_CONTROLLER_URL': 'quad_plus_sitl'},
```
ad-daniel, line 39:
```suggestion
        additional_env={'WEBOTS_CONTROLLER_URL': controller_url_prefix() + 'quad_plus_sitl'},
```

ad-daniel: same

ad-daniel (changes requested):

src/opendr/planning/end_to_end_planning/__init__.py
halil93ibrahim: Yes, I switched it with ROS_VERSION check, instead of ROS_DISTRO.
The reason for this is, the env (UAV_depth_planning_env) relies on rospy, hence, ROS1. It is not necessary for inference, so ROS2 node can run without importing it.

src/opendr/planning/end_to_end_planning/dependencies.ini
```diff
@@ -14,4 +14,3 @@ python=vcstool
 #  https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format
 python=stable-baselines3
 linux=ros-noetic-webots-ros
```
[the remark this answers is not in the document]
halil93ibrahim: I also added noetic-webots-ros to install.sh instead of here. Is that correct?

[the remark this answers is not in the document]
ad-daniel: yes, and the patch mentioned above we can also put it in install.sh temporarily (for ros2 only)

projects/opendr_ws_2/src/opendr_planning/opendr_planning/end_to_end_planning_robot_driver.py
```diff
@@ -0,0 +1,25 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import rclpy
+
+
+class EndToEndPlanningUAVRobotDriver:
+    def init(self, webots_node, properties):
+        rclpy.init(args=None)
+        self.__node = rclpy.create_node('end_to_end_planning_uav_robot_driver')
```
tsampazk, line 22: This class seems to be used as `plugin type` [here](https://github.com/opendr-eu/opendr/blob/778f36698447911772cdb85dcd04753aee6216fc/projects/opendr_ws_2/src/opendr_planning/resource/uav_robot.urdf#L32). I am not exactly familiar to how this is used, but both arguments seem unused and init is missing the underscores. If they are supposed to be this way, feel free to ignore this and resolve the comment.

tsampazk (commented):
Thanks! I have just one comment below. 

Moreover, maybe we need to update the webots version stated in the README [here](https://github.com/opendr-eu/opendr/blob/ros2/docs/reference/end-to-end-planning.md#simulation-environment-setup)?

src/opendr/planning/end_to_end_planning/dependencies.ini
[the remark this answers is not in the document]
ad-daniel: please note that I changed the temporary fix above as this is better

projects/opendr_ws_2/src/opendr_planning/opendr_planning/end_to_end_planning_robot_driver.py
ad-daniel: no, it has to be init without underscores otherwise the plugin is not built correctly

src/opendr/planning/end_to_end_planning/__init__.py
```diff
@@ -1,4 +1,6 @@
 from opendr.planning.end_to_end_planning.e2e_planning_learner import EndToEndPlanningRLLearner
-from opendr.planning.end_to_end_planning.envs.UAV_depth_planning_env import UAVDepthPlanningEnv
+import os
+if os.environ.get("ROS_VERSION") == "1":
```
ad-daniel, line 3:
```suggestion
if os.environ.get("ROS_DISTRO") == "melodic" or os.environ.get("ROS_DISTRO") == "noetic":
```

ad-daniel: I don't believe ROS_VERSION is present out of the box, and we don't want to add another environmental variable considering it can be inferred from ROS_DISTRO. Moreover, we only really support melodic and noetic so I think this is better.

ad-daniel (changes requested):
Last detail, otherwise works for me, thanks!

ad-daniel:
I'll take a look at the failures, but if it's "rate-limit" or single demonstration grasping then they can be ignored

projects/opendr_ws_2/src/opendr_planning/opendr_planning/end_to_end_planning_robot_driver.py
halil93ibrahim: Then resolving this comment without action.

ad-daniel:
the failure is unrelated

ad-daniel (approved):

tsampazk (approved):
Thank you.

Issue #362, opened 2022-11-29
BoundingBoxList(Target) ctor throws RuntimeWarning when provided with empty list


tsampazk:
I noticed that when running various object detection inference demos i got the following warning:

```
/opendr/venv/lib/python3.8/site-packages/numpy/core/fromnumeric.py:3474: RuntimeWarning: Mean of empty slice.
  return _methods._mean(a, axis=axis, dtype=dtype,
/opendr/venv/lib/python3.8/site-packages/numpy/core/_methods.py:189: RuntimeWarning: invalid value encountered in double_scalars
  ret = ret.dtype.type(ret / rcount)
```

This is caused by the corresponding learners' infer method, which initialize a `BoundingBoxList` object providing and empty list `[]`, e.g. [here](https://github.com/opendr-eu/opendr/blob/cf99c7b398ef8e9b0ffe6fcb0f13a16c593f467f/src/opendr/perception/object_detection_2d/centernet/centernet_learner.py#L436). This leads to the [constructor](https://github.com/opendr-eu/opendr/blob/cf99c7b398ef8e9b0ffe6fcb0f13a16c593f467f/src/opendr/engine/target.py#L444) which tries to extract confidence from the boxes but fails as `data` is empty.

Various simple fixes can be applied such as checking for the length of `self.data` but this leads to a situation where appending boxes to data (which as far as i can tell is a common usage), leaves `self.confidence` empty as it doesn't update.

I suggest to add a method similar to append for the `BoundingBoxList` class which would also update the `self.confidence` attribute. Another possibility is to force the user to provide a non-empty list to the ctor. Please let me know how does these solutions sound and whether there might be implications to consider. @iliiliiliili

PR #365, opened 2022-11-30
Fix bbox/tracking 2D/3D list confidence

iliiliiliili:
- Fix `BoundingBoxList`, `TrackingAnnotationList` `BoundingBoxList3D`, `TrackingAnnotationList3D` confidence warnings on empty lists #362 
- Fix undefined `image_id` and `segmentation` for coco `BoundingBoxList`


src/opendr/engine/target.py
```diff
@@ -437,11 +437,13 @@ class BoundingBoxList(Target):
     """
     def __init__(
         self,
-        boxes,
+        boxes=[],
```
tsampazk, line 440: Default value shouldn't be a mutable value, i would suggest defaulting to `None` and inside the method check for `None` and assign empty list if so.

tsampazk (commented):
Thanks Illia for the fixes, changes look good, i have just one comment.

src/opendr/engine/target.py
iliiliiliili: Fixed

iliiliiliili:
Some strange errors occur across different tests with no module found, such as:
`ModuleNotFoundError: No module named 'cv2'`

tsampazk:
> Some strange errors occur across different tests with no module found, such as: `ModuleNotFoundError: No module named 'cv2'`

There seems to be some wider issue at play here, i don't think the errors in the tests are related to the changes in this PR.

ad-daniel:
I've re-run the PR and the problems seem to have gone. Likely a CI hiccup

tsampazk (approved):
Thanks!

PR #317, opened 2022-09-26
Continual Transformer Encoder

LukasHedegaard:
New tool for efficient application of Transformer Encoders to time-series data.
Note: this does not come with pre-trained models. The user must fit the learner on her own features.

PR #369, opened 2022-12-06
Fix sklearn dependency

ad-daniel:
sklearn was deprecated in favor of scikit-learn, which now results in an error


ad-daniel:
Easier to test/address it directly in the PR where it's broken

ad-daniel:
no longer needed as #317 will be merged first and includes the fix

PR #373, opened 2022-12-08
ROS1 - Minor fixes

tsampazk:
Found a couple of small issues with two nodes:
1. A duplicate shortcut on object tracking 2d fair mot 
2. The visualization of panoptic segmentation didn't show on rqt due to bad encoding and threw this error:
    ```
    ImageView.callback_image() could not convert image from '8UC3' to 'rgb8' ([8UC3] is not a color format. but [rgb8] is. The conversion does not make sense)
    ```


ad-daniel (approved):
nice catch, thank you

passalis (approved):
Thank you!

PR #375, opened 2022-12-12
Fix nightly tests in CI

ad-daniel:
Ambiguity measure test failed in the nightly [here](https://github.com/opendr-eu/opendr/actions/runs/3671296284/jobs/6206537808), the reason is that although created, the package wasn't installed prior to running the test

EDIT: yolov5 was also failing when run on its own


ad-daniel:
I also fixed the docker tests due to the change made this morning. It's only run when doing `test release` label hence why it wasn't caught.

stefaniapedrazzi (approved):

tsampazk (approved):
Thank you.

PR #256, opened 2022-05-10
ROS2 workspace and example nodes

tsampazk:
This PR contains a new ROS2 (**Foxy-Fitzroy** version) workspace located in the `projects` directory and, for now, it serves to ~test and finalize the structure, naming, etc.,~ gather and finalize all ROS2 nodes in a unified PR. Right now there are no ~docstrings~ (docstrings added), documentation or READMEs. This description will get updated for any additions.

Contents:

1. `opendr_perception` python package
   - Contains a pose estimation node
   - Contains fall detection node
   - Contains object detection 2d centernet/detr/ssd/yolov3 nodes
   - Contains face detection retinaface node
   - Contains face recognition node
   - Contains semantic segmentation bisenet node
   - ~Contains a subscriber tester node (tester), that subscribes to the messages published by the pose estimation node for 
testing~ Testing can be performed as described in steps 9 and 10 of **Building and Running** below.
2. `opendr_ros2_bridge` python package
   - Contains `bridge.py` which includes a class with methods to convert images, poses, etc. from and to ROS2 messages 
   - This uses [`cv_bridge`](https://github.com/ros-perception/vision_opencv/tree/ros2/cv_bridge) which is included in the `vision_opencv` package
3. `opendr_ros2_messages` CMake package
   - Contains custom OpenDR messages, right now only `OpenDRPose2D` and `OpenDRPose2DKeypoint`
   - Note that this package is a *CMake package* due to limitations in ROS2: [custom messages cannot be generated in python packages](https://docs.ros.org/en/foxy/Tutorials/Beginner-Client-Libraries/Custom-ROS2-Interfaces.html#create-a-new-package)

The logic behind the structuring of the packages and nodes is similar to OpenDR's ROS1 packages/nodes.

Below you can find instructions to install, build and run the nodes for testing. Note that i did everything on a system with ROS1 already installed.

_I faced many issues along the way that might reappear in a fresh install of ROS2, etc., so if any problems/errors occur following the instructions please get in touch with me, to possibly save you some time._

Installation
====
- To install ROS2 i followed [this](https://programmer.ink/think/installation-tutorial-of-ros1-and-ros2.html) tutorial (section 2), which installs the 'foxy' release of ROS2. (Note that on '(7) configure environment variables', you need to replace `dashing` with `foxy`)
- Edit: At this point you might need to run `sudo apt-get install ros-foxy-vision-msgs` as discussed [below](https://github.com/opendr-eu/opendr/pull/256#issuecomment-1161951250)
[...]

PR #376, opened 2022-12-12
Improve installation script and docs

ad-daniel:
This PR does:
- Since ros2 interface is in the works we no longer can impose the ROS_DISTRO variable, but rather we should let the user set it and install the additional dependencies accordingly. Additionally, it's pointless to install dependencies for ROS if the user wants to use ROS2, and vice-versa.
- rather than compiling from source, we can install the usb-cam ROS/ROS2 package instead. Not all tools need it it's true, but perception is most of what the opendr tools do so I don't think it's a problem (especially since it's a small utility).
- some doc changes (primarily to let the user know that it has to set the environmental variables before installing)

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -40,7 +40,7 @@ make install_compilation_dependencies
 make install_runtime_dependencies
 
 # Install additional ROS packages
-if [[ ${ROS_DISTRO} == "noetic" || ${ROS_DISTRO} == "melodic"]]; then
+if [[ ${ROS_DISTRO} == "noetic" || ${ROS_DISTRO} == "melodic" ]]; then
   echo "Installing ROS dependencies"
   sudo apt-get -y install ros-$ROS_DISTRO-vision-msgs ros-$ROS_DISTRO-geometry-msgs ros-$ROS_DISTRO-sensor-msgs ros-$ROS_DISTRO-audio-common-msgs ros-$ROS_DISTRO-usb-cam
 fi
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -9,21 +9,21 @@ if [[ -z "${OPENDR_DEVICE}" ]]; then
   export OPENDR_DEVICE=cpu
 fi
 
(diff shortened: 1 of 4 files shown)


stefaniapedrazzi (approved):

bin/install.sh
```diff
@@ -40,7 +40,16 @@ make install_compilation_dependencies
...
-sudo apt-get install ros-$ROS_DISTRO-vision-msgs ros-$ROS_DISTRO-geometry-msgs ros-$ROS_DISTRO-sensor-msgs ros-$ROS_DISTRO-audio-common-msgss
+if [[ ${ROS_DISTRO} == "noetic" || ${ROS_DISTRO} == "melodic" ]]; then
+  echo "Installing ROS dependencies"
+  sudo apt-get -y install ros-$ROS_DISTRO-vision-msgs ros-$ROS_DISTRO-geometry-msgs ros-$ROS_DISTRO-sensor-msgs ros-$ROS_DISTRO-audio-common-msgs ros-$ROS_DISTRO-usb-cam
+fi
+
+# Install additional ROS2 packages
+if [[ ${ROS_DISTRO} == "foxy" || ${ROS_DISTRO} == "humble" ]]; then
+  echo "Installing ROS2 dependencies"
+  sudo apt-get -y install ros-$ROS_DISTRO-usb-cam
```
tsampazk, line 51: I think that we might need to add some more installation commands here as described in the `Installation` section in the description of #256?

tsampazk: `vision-msgs` and `colcon` probably, and some additional similar to ROS1 above when more ROS2 nodes are added .

ad-daniel: yes, this PR targets develop, when merging it into `ros2` branch we can adapt it

tsampazk: Oh alright, thanks!

tsampazk (approved):
Thanks, installation of ROS looks much better now.

PR #378, opened 2022-12-14
SiamRPN fixes and ab3dmot ros1 node duplicate shortcut

tsampazk:
Very minor fix for siamrpn node path. 

It wasn't caught as by using `catkin_make` the scripts are not installed, i.e. only the `devel` directory is created and needs to be sourced, so this didn't produce any issue.

This is an thing that was brought up before, some nodes, but not all, are in the `CMakeLists.txt` `catkin_install_python(PROGRAMS ... )` for ROS1. As far as i can tell, these are used when you do `catkin_make install` and source the `install/setup.bash` to run the nodes.

Edit: Integrated another minor fix that i missed earlier for object tracking 3d ab3dmot ROS1 node. The duplicate shortcut in argparse stopped the node from running entirely.


ad-daniel (dismissed):
nice catch, thank you

passalis (approved):
Thank you!

ad-daniel (dismissed):

tsampazk:
I added another fix on this PR, for siamrpn ros1 node to include a built-in detector to initialize the tracker, so as not to open a different PR.

Now the node initializes the tracker using an object detector. Out of the bboxes detected, the one closer to the input image center is chosen for initialization.

We've discussed this with @vivinousi, and concluded that it is the easiest way to initialize the tracker and use the node out-of-the-box.

I've already made this change in ROS2.

passalis (approved):
Thank you!

ad-daniel (approved):
Thank you both, looks good to me as well!

PR #379, opened 2022-12-14
ROS1 node fix duplicate argparse shortcut tracking3d

tsampazk:
Another minor fix that i missed earlier for object tracking 3d ab3dmot ROS1 node. This stopped the node from running entirely.


tsampazk:
I think it's better to integrate this in the earlier PR #378, sorry.

PR #381, opened 2022-12-19
Debug test suite develop

ad-daniel:
Looks like something related to numpy broke the test suite very recently, this branch is for debug

### DIFF
--- a/.github/workflows/tests_suite.yml
+++ b/.github/workflows/tests_suite.yml
@@ -106,6 +106,7 @@ jobs:
         make install_compilation_dependencies
         make install_runtime_dependencies
         pip install -r tests/sources/requirements.txt
+        pip list
         if [ ${{ matrix.package }} = "ctests" ]; then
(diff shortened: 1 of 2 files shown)


ad-daniel:
It seems that the difference between yesterday (when things worked) and and today is that numpy 1.24.0 is installed instead of 1.23.5. With 1.24.0 (and 1.25.0) several deprecated functionalities have been removed, hence why the failures. As a temporary solution I think we can just restrict the upper version to be installed, I'll open an issue about the deprecation warnings so that we can address them when things calm down a bit after the release, there's several more beyond numpy that we should fix anyway so we should address them all at the same time.

The restriction worked, I just forgot the test sources label

passalis (approved):
Thanks for working on this!

tsampazk (approved):
Thanks Daniel!

stefaniapedrazzi (approved):

PR #385, opened 2022-12-20
ROS1 readme additions

tsampazk:
Added single object tracking 2d SiamRPN documentation as well as index entry. 

Also added nanodet entry in object detectors 2d.


passalis (approved):
Thank you!

ad-daniel (approved):
Thank you!

PR #386, opened 2022-12-21
Test whether single-demo-grasp is broken in the CI 

ad-daniel:
It was reported by a partner and it failed in the nightly, so it might be broken


ad-daniel:
As far as I can tell, every time it failed 1.24.0 was installed, why does it show only now and not in the previous PR where it was fixed is beyond me. Either way the upper limit must be specified everywhere since the tools might be installed separately.

tsampazk (approved):
Thanks Daniel!

PR #390, opened 2022-12-30
Fix package creator and sources

ad-daniel:
The [creation](https://github.com/opendr-eu/opendr/actions/runs/3807895638/jobs/6478014998) was successful, however the root package wasn't uploaded due to a missing new line in `packages.txt`.
I've uploaded the missing one manually, this is just to ensure everything is fine


ad-daniel:
mmh, actually running the `test release` wouldn't really test it. In the workflow we install ALL the packages (1 by 1) and then test that all unittests still pass. We probably did it this way to ensure it used the locally generated packages rather than install the root and download its dependencies from pypi). I see benefits in both approaches, but perhaps we should still manually ensure that by installing it from the root package everything still works

ad-daniel:
It seems that the following two installation methods:
- install engine + opendr-toolkit (full install)
- install engine + each pip package separately

although identical in theory,  they do not yield the same result dependency wise and the first approach has dependency issues whereas the second doesn't. This PR only addresses:

- fix copyright (& test sources)
- version bump
- fix package publisher not creating a package for the entire toolkit (missing new line)

tsampazk (approved):
Thank you Daniel!

passalis (approved):
Thank you!

PR #397, opened 2023-01-16
Fall Detection - alternative infer input

### LINKED ISSUE #282: Fall Detection - Provide infer method that works with either image or pose
### Current situation

Edit: Main issue is fixed see comment below.

Right now, the standalone fall detection tool [uses](https://github.com/opendr-eu/opendr/blob/3a9a119f2f4391c3d9ba5f2f998400ffb61f72cf/src/opendr/perception/fall_detection/fall_detector_learner.py#L219-L228) a pose estimation learner internally ([which is provided in the learners ctor](https://github.com/opendr-eu/opendr/blob/3a9a119f2f4391c3d9ba5f2f998400ffb61f72cf/src/opendr/perception/fall_detection/fall_detector_learner.py#L30-L33)). The inference method requires an image. Inference also [returns the pose for further usage](https://github.com/opendr-eu/opendr/blob/3a9a119f2f4391c3d9ba5f2f998400ffb61f72cf/src/opendr/perception/fall_detection/fall_detector_learner.py#L314-L330), which might be counterintuitive.

### New feature
It would be helpful to be able to run fall detection inference with provided OpenDR pose instead of image, by running a pose estimator externally. This usage is closer to the described usage of "cheap fall detection when pose estimation is already performed for other purposes" and can "disentangle" the used pose estimation learner from the fall detection learner, if the user chooses.

### Implementation
Out of various options (e.g. multiple infer methods with different names) and because overloading is not possible in a straightforward manner in python, i would suggest putting image and pose both as optional arguments with `None` as default for the existing infer method and managing them internally in the method. 

### ROS implications
This can also improve the fall detection ROS node, giving the option to run standalone and take the pose from a ROS message. Right now in ROS, the pose gets consumed and cannot be used for other purposes as it is not published by the node.

Any suggestions are welcome.

tsampazk:
This PR basically fixes #282, by changing the infer method of the fall detection learner. **The infer method now accepts pose lists as input in addition to OpenDR or OpenCV images.** I also simplified the return by dropping the "detection lines".

Along with this change i have fixed and/or improved the demos, the ROS1/2 nodes, the jupyter notebook tutorial and the docs, and added additional tests in the test_infer method.

I will keep the issue open to investigate further on how the ROS1/2 nodes can provide an alternate mode where they will subscribe to a pose annotation topic instead of an image topic so they can be chained with a pose estimator node.


ad-daniel (approved):
Looks good to me, thank you!

Pavlos-Tosidis (dismissed):
Looks good for me

ad-daniel:
the CI seems to fail the pose estimation test, I'll rerun it once the other tests are done but does it seem legitimate @tsampazk or a fluke?

tsampazk:
Hmm, that's weird @ad-daniel. The pose estimation test failed on a high resolution pose estimator test which is quite unrelated to this PR. The most likely culprit is #409 that changed some stuff in the hr pose estimator learner. As far as i can see the pose estimation test didn't run before that PR was merged. Give me some time to look into it.

tsampazk:
Ok the hr pose estimation test wasn't updated according to the latest changes in the infer method from #409. Should be good to go now.

ad-daniel (approved):
That did the trick, thank you Kostas! You can safely ignore the canceled test, it will be gone in the other PR

ad-daniel:
@Pavlos-Tosidis sorry to bother again, we need the second approval here too

Pavlos-Tosidis (approved):
There you go!

PR #359, opened 2022-11-28
New feature: LiDAR-based panoptic segmentation - EfficientLPS

aselimc:
Hi,

This PR adds [EfficientLPS](http://lidar-panoptic.cs.uni-freiburg.de/) network. The original repo can be found [here](https://github.com/robot-learning-freiburg/EfficientLPS). 

ROS Nodes are also added together with this PR. Furthermore, we have introduced PointCloud2 publisher, which is the latest PointCloud message type in ROS.

This PR does not need to be merged before the next release.

[Edit] Performance evaluation has to be completed but I don't have a recent Jetson right now. @vniclas

PR #404, opened 2023-02-03
ROS2 Node for EfficientLPS

aselimc:
Hi,

This PR adds [EfficientLPS](http://lidar-panoptic.cs.uni-freiburg.de/) ROS2 nodes. It should be merged after the #359. It simply adds the same functionality as we added for ROS nodes but this is for ROS2.


tsampazk (commented):
Thank you for the new node! We are missing a ROS2 dependency for the new messages used, i added it in our build file.

When running the point cloud publisher node without the `-t` flag it produces the following error:

```
opendr_perception/point_cloud_2_publisher_node.py", line 65, in start
    point_cloud = self.dataset[i % len(self.dataset)][0]
AttributeError: 'PointCloud2DatasetNode' object has no attribute 'dataset'
```
which is not particularly useful, instead of getting the error in line 80, similar to the ROS1 node. Could you please look into this so it properly prints the correct error message in case a dataset is not provided and the flag `-t` not used?

aselimc:
> Thank you for the new node! We are missing a ROS2 dependency for the new messages used, i added it in our build file.
> 
> When running the point cloud publisher node without the `-t` flag it produces the following error:
> 
> ```
> opendr_perception/point_cloud_2_publisher_node.py", line 65, in start
>     point_cloud = self.dataset[i % len(self.dataset)][0]
> AttributeError: 'PointCloud2DatasetNode' object has no attribute 'dataset'
> ```
> 
> which is not particularly useful, instead of getting the error in line 80, similar to the ROS1 node. Could you please look into this so it properly prints the correct error message in case a dataset is not provided and the flag `-t` not used?

Sorry for this, I will fix it probably later today (studying to exams 😄). I was wondering that the error message in line 80 is still produced before `AttributeError` right?

tsampazk:
> Sorry for this, I will fix it probably later today (studying to exams smile). I was wondering that the error message in line 80 is still produced before `AttributeError` right?

No problem, i think it doesn't print at all :confused:

aselimc:
> > Sorry for this, I will fix it probably later today (studying to exams smile). I was wondering that the error message in line 80 is still produced before `AttributeError` right?
> 
> No problem, i think it doesn't print at all 😕

This should be addressed now. Thanks for noticing this 😄

tsampazk (approved):
Thank you!

ad-daniel (approved):
Thank you!

PR #409, opened 2023-02-14
High Resolution Pose Estimation webcam demo

mthodoris:


tsampazk (changes requested):
Thank you for the new demo, it is really useful for comparing the two methods.

I have some mostly minor comments on the code itself. I tested the demo locally and noticed that it ran ok for webcam, but when i used an ip camera via the android app DroidCam, providing high resolution input, i got two issues:
1. The images came out with a distorted aspect ratio, in my case they appeared wider
2. There was a significant delay similar to what we had in ROS #275, and as far as i could tell it got worse and worse while it was running. To clarify a bit, i had a smooth framerate a delay between when e.g. a hand move was made and when i saw the move on the output.

tsampazk (approved):
Thank you, it works well now with just the webcam.

PR #410, opened 2023-02-14
fix(efficientLPS): Fix the dependency conflict of geffnet installation

aselimc:
This PR aims to fix the dependency error between EfficientPS and EfficientLPS. EfficientLPS is unable to use efficientNet (geffnet) library that can be installed using EfficientPS even though the versions are same. The reason is because original efficientNet library is modified for EfficientLPS. However, EfficientPS is still compatible with the modified version. That is why I have added a small trick when `python_dependencies.txt` file is being created by `dependencies/parse_dependencies.py`.


bin/install_nvidia.sh
```diff
@@ -302,7 +302,8 @@ pip3 install git+https://github.com/mapillary/inplace_abn.git
 pip3 install git+https://github.com/facebookresearch/detectron2.git@4841e70ee48da72c32304f9ebf98138c2a70048d
 pip3 install git+https://github.com/cidl-auth/DCNv2
 pip3 install ${OPENDR_HOME}/src/opendr/perception/panoptic_segmentation/efficient_ps/algorithm/EfficientPS
-pip3 install ${OPENDR_HOME}/src/opendr/perception/panoptic_segmentation/efficient_ps/algorithm/EfficientPS/efficientNet
+pip3 install ${OPENDR_HOME}/src/opendr/perception/panoptic_segmentation/efficient_lps/algorithm/EfficientLPS
```
ad-daniel, line 305: I don't believe this script for embedded platforms relies on `parse_dependencies.py` as the dependencies are all hard-coded, so I suspect the issue there would persist. So I think the other solution would need to be implemented in this case

ad-daniel (changes requested):
Thank you, just a doubt

ad-daniel:
On a side note, I've noticed that whenever installing the toolkit a diff is created in the EfficientLPS submodule, specifically:
`opendr/src/opendr/perception/panoptic_segmentation/efficient_lps/algorithm/EfficientLPS`

b

It's not a big deal, but rather an annoyance (being a submodule I need to cd into it to clear it every time I switch branches). Any chance we can get rid of this?

tsampazk:
> On a side note, I've noticed that whenever installing the toolkit a diff is created in the EfficientLPS submodule

I can confirm that I've been having this issue too pretty randomly on and off in the past months.

bin/install_nvidia.sh
aselimc: I have switched this to EfficientLPS because the reason I mentioned as EfficientLPS can in fact not be able to use geffnet installation from EfficientPS submodule and be on the safe side. Unfortunately I am not an expert of how this script is used. I had a look at this because this was mentioned in the issue 😅.

What would be your suggestion here ?

ad-daniel: Sorry I was under the impression that both `efficientNet` were being installed again but I must have read it wrong, my bad!

ad-daniel (approved):
Thank you. The failure is due to the presence of parenthesis in the branch name, but since `test tools` works, docker is bound to work as well

tsampazk (approved):
Thank you!

aselimc:
> On a side note, I've noticed that whenever installing the toolkit a diff is created in the EfficientLPS submodule, specifically: `opendr/src/opendr/perception/panoptic_segmentation/efficient_lps/algorithm/EfficientLPS`
> 
> b
> 
> It's not a big deal, but rather an annoyance (being a submodule I need to cd into it to clear it every time I switch branches). Any chance we can get rid of this?

Ah sorry I totally forgot to respond to this. Well, this happens randomly in my setup as well, but I think the reason is because when submodule packages are installed, they tend to create this `version.py` and/or modify existing packaging info etc. I have not figured out how to prevent this, but as you have written this is not a big deal, so I would suggest we merge this and handle the other problem separately.

Issue #414, opened 2023-02-28
Nanodet C api fails on docker


ad-daniel:
During the [building](https://github.com/opendr-eu/opendr/actions/runs/4287737831/jobs/7468944494) of the docker, a failure occurs when compiling the C api:

```
Processing triggers for libc-bin (2.31-0ubuntu9.9) ...
make[1]: Entering directory '/opendr/src/c_api'
Building C API...
g++ -fPIC -c opendr_utils.cpp -o /opendr/build/opendr_utils.o -I/usr/local/include/onnxruntime/ -I/usr/local/include/rapidjson/ `pkg-config --cflags opencv4` -I/opendr/include
g++ -fPIC -c activity_recognition_x3d.cpp -o /opendr/build/opendr_x3d.o -I/usr/local/include/onnxruntime/ -I/usr/local/include/rapidjson/ `pkg-config --cflags opencv4` -I/opendr/include
g++ -fPIC -c face_recognition.cpp -o /opendr/build/opendr_face_recognition.o -I/usr/local/include/onnxruntime/ -I/usr/local/include/rapidjson/ `pkg-config --cflags opencv4` -I/opendr/include
g++ -fPIC -c lightweight_open_pose.cpp -o /opendr/build/opendr_open_pose.o -I/usr/local/include/onnxruntime/ -I/usr/local/include/rapidjson/ `pkg-config --cflags opencv4` -I/opendr/include
g++ -fPIC -c object_detection_2d_detr.cpp -o /opendr/build/opendr_detr.o -I/usr/local/include/onnxruntime/ -I/usr/local/include/rapidjson/ `pkg-config --cflags opencv4` -I/opendr/include
g++ -fPIC -c object_detection_2d_nanodet_jit.cpp -o /opendr/build/opendr_nanodet_jit.o -I/usr/local/include/onnxruntime/ -I/usr/local/include/rapidjson/ `pkg-config --cflags opencv4` -I/opendr/include -I/usr/local/libtorch/include -I/usr/local/libtorch/include/torch/csrc/api/include
object_detection_2d_nanodet_jit.cpp:18:10: fatal error: torch/script.h: No such file or directory
   18 | #include <torch/script.h>
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
make[1]: Leaving directory '/opendr/src/c_api'
make[1]: *** [Makefile:49: /opendr/lib/libopendr.so] Error 1
make: *** [Makefile:56: libopendr] Error 2
[...]

PR #418, opened 2023-03-01
Docker test release

ManosMpampis:
This PR have fixes about docker compilation on PC for CUDA and CPU.
It was made after Issue [Nanodet C api fails on docker](https://github.com/opendr-eu/opendr/issues/414).
It was tested in CUDA 11.2


ManosMpampis:
@ad-daniel  Can you double check this PR? I have tested with "test release" and docker was build successfully. I whould like to be tested in CUDA from someone else too before merge this PR.

ad-daniel (approved):
The build process worked for me and appears to be working in the CI as well, I no longer see the compilation error. I don't know how generic the solution you propose is (whether we can always find a libtorch for download based on the CUDA_VERSION determined with this method (seems a bit a hack)), but I think we can proceed this way for now and if problems arise we can either fix them or revert this PR.

The failures in the CI are unrelated

stefaniapedrazzi (approved):
Thank you!

ManosMpampis:
It is ready for review, in the current status the 1.9.0 torch is installing for C implementation. If we want 2 different implementations, one for the embededs and one for the pc, we can first install the runtime dependences and much the version of pytorch in dependences/download_torch.py

PR #419, opened 2023-03-13
ROS nodes FPS performance measurements

tsampazk:
This PR adds time performance measurement of tools' inference. For each node, the time it takes to run (only) inference is measured and published in a performance topic. Publishing this message is optional, i.e. the relevant topic needs to be set via argparse. This message can be subscribed to, or echoed to show the current FPS.

If a more accurate performance measurement is needed, one can use the new performance node which subscribes to the performance topic and calculates a running average of the FPS measurement which it prints in the console along with the time it took for the last frame to be processed.

- [x] ROS1 Pose estimation node proof of concept
- [x] ROS1 Performance node
- [x] ROS1 Documentation update
- [x] ROS1 Other nodes
- [x] ROS2 Nodes
- [x] ROS2 Documentation update

@thomaspeyrucain @ad-daniel Let me know how this looks and whether it's sufficient and/or convenient, before i add it in the rest of the ROS1/2 nodes.


ad-daniel:
Implementation wise I think this is a nice way of doing it. I have some doubts concerning whether measuring just the infer method is enough though, because in that case I don't expect there to be any real difference between the value obtained from this method and the values already reported in the documentation when the tools were benchmarked using the python API. It's measuring the same thing. So the question is, can we actually use these values to "show" that the performance requirements of the use-case have been met? Because by taking only the infer method, it sounds to me like it's an evaluation of the tool, not a task/use-case. What else should be included and where the boundary should be I don't know.. but before doing the change for all nodes it should be settled

tsampazk:
> Implementation wise I think this is a nice way of doing it. I have some doubts concerning whether measuring just the infer method is enough though, because in that case I don't expect there to be any real difference between the value obtained from this method and the values already reported in the documentation when the tools were benchmarked using the python API. It's measuring the same thing. So the question is, can we actually use these values to "show" that the performance requirements of the use-case have been met? Because by taking only the infer method, it sounds to me like it's an evaluation of the tool, not a task/use-case. What else should be included and where the boundary should be I don't know.. but before doing the change for all nodes it should be settled

Oh i was under the impression that that was what we wanted (measure only tool inference). I'll wait for more input before moving forward.

projects/opendr_ws_2/src/opendr_perception/opendr_perception/object_tracking_3d_ab3dmot_node.py
```diff
@@ -78,19 +88,26 @@ def callback(self, data):
         :param data: input message
         :type data: sensor_msgs.msg.Image
         """
-
+        if self.performance_publisher:
+            start_time = perf_counter()
         # Convert sensor_msgs.msg.Image into OpenDR Image
         point_cloud = self.bridge.from_ros_point_cloud(data)
         detection_boxes = self.detector.infer(point_cloud)
+        tracking_boxes = self.learner.infer(detection_boxes)
```
stefaniapedrazzi, line 96: Would it maybe make sense here to execute it only when needed?
```suggestion
        if self.tracking_id_publisher or self.performance_publisher:
            tracking_boxes = self.learner.infer(detection_boxes)
```

tsampazk: Thank you for the review Stefania! 

The reasoning behind this was that this node is supposed to do tracking so it is a bit counterintuitive to disable it. It runs the Voxel3D object detection tool, so one should use that tool's node instead of this one. Nevertheless i can imagine someone wanting to have both options available when using this so i will add your suggestion and apply it to the ROS1 node too.

stefaniapedrazzi: Ok, thank you for the explanation!
I just checked the code and saw that it was used only if `self.tracking_id_publisher` was set.
But for me it is also fine to keep it without the if condition.

stefaniapedrazzi (approved):

projects/opendr_ws_2/src/opendr_perception/opendr_perception/object_tracking_3d_ab3dmot_node.py
tsampazk: Yeah i think code-wise it makes more sense with the if :smile: Thanks!

projects/opendr_ws/src/opendr_perception/scripts/performance_node.py
```diff
@@ -0,0 +1,76 @@
...
+        rospy.spin()
+
+    def callback(self, data):
+        """
+        Callback that processes the input data and publishes to the corresponding topics.
+        :param data: Input image message
+        :type data: sensor_msgs.msg.Image
+        """
+        fps = data.data
+        rospy.loginfo(f"Time per frame: {str(round(1.0 / fps, 4))} sec")
```
passalis, line 54:
```suggestion
        rospy.loginfo(f"Time per inference: {str(round(1.0 / fps, 4))} sec")
```

projects/opendr_ws/src/opendr_perception/scripts/performance_node.py
```diff
@@ -0,0 +1,76 @@
...
+        while len(self.fps_window) < self.window_length:
+            self.fps_window.append(fps)
+        self.fps_window = self.fps_window[1:]
+        self.fps_window.append(fps)
+        rospy.loginfo(f"Average FPS   : {round(mean(self.fps_window), 2)}")  # NOQA
```
passalis, line 59:
```suggestion
        rospy.loginfo(f"Average inferences per second   : {round(mean(self.fps_window), 2)}")  # NOQA
```

passalis (commented):
Thank you for all this extensive work! I have only two minor suggestions for the performance node (switch from FPS to inferences/sec since some nodes do not receive frames as input).

passalis (approved):
Thank you!

PR #426, opened 2023-04-04
Fix: EfficientLPS panoptic segmentation coloring bug

aselimc:
EfficientLPS panoptic segmentation tool had a small bug in the panoptic/instance coloring. With this PR, it is aimed to be fixed.


tsampazk (approved):
I ran the ROS node, which i as far as i can tell would verify that the colors are indeed ok. To verify, this is what i get:

image

If this looks right to you, the code changes look good to me! Thank you.

vniclas:
Thanks @tsampazk !

PR #433, opened 2023-05-12
Speech Transcription with Whisper and Vosk

minhquoc0712:
This branch integrates two speech transcription libraries: Whisper by openAI and Vosk.

**Features**:
- Adding target classes for the output of each transcription library.
- Integrate the download, load, infer and eval, etc, of the two libraries into OpenDR learner classes.
- Add a demo for the speech transcription task.
- Implement ROS  node for the speech transcription task with two backbones, Whisper and Vosk.
- Implement ROS2 node for the speech transcription task.
- Add document for ros bridge, ros messages, transcription target, and learner class.

**Notes**:
- For ROS node, transcribing streaming speech is not native for Whisper. Therefore I add some heuristic rules for breaking the phrases, detecting silence, etc.
- For RO2, the audio common package seems to update the default audio topic to `audio/audio`. If that is the case, I think we need to change the default input audio topic for other audio-related ROS2 nodes.

**Known issues**:
- Currently, there is a problem with the ROS2 node. It cannot do the speech transcription properly. The reason could be the audio format published by the audio capture node in ROS2.


src/opendr/perception/speech_transcription/vosk/vosk_learner.py
```diff
@@ -0,0 +1,382 @@
...
+        transcriptions = pd.DataFrame(dict(hypothesis=hypotheses, reference=references))
+        if save_path_csv is not None:
+            transcriptions.to_csv(save_path_csv, index=False)
+
+        wer = jiwer.wer(list(transcriptions["reference"]), list(transcriptions["hypothesis"]))
+
+        return {"wer": wer}
+
+    def optimize(self):
+        return
```
stefaniapedrazzi, line 379: For methods that are not implement usually this is used (see for example the efficient_ps_learner.py).
Could you please update all the occurrences in your new learners?
```suggestion
        """This method is not used in this implementation."""
        raise NotImplementedError
```

stefaniapedrazzi (dismissed):
Thank you.
Here are some initial comments.
I also added the new learners to the test suite.

Changes to src/opendr/perception/panoptic_segmentation/efficient_ps/algorithm/EfficientPS`seems unrelated and should be reverted.

minhquoc0712:
Hi, can you review my pull request? @omichel , @passalis ?

minhquoc0712:
Hi, I have fixed the styles

passalis (approved):
Thank you!

PR #436, opened 2023-05-16
Add RGB gesture recognition

katerynaCh:
This PR implements RGB-based hands gesture recognition (detection) based on [hagrid dataset](https://github.com/hukenovs/hagrid/tree/master) for 18 hand gestures. 

We currently provide one pretrained model. ROS1/ROS2 nodes are included, as well as a webcam demo. 

Algorithmically it relies on existing implementation of nanodet (I thought it is better from application perspective to create a dedicated learner for gesture recognition, but please let me know if you think otherwise and I change it, and just provide necessary model/dataset processing/configs as part of nanodet learner).


stefaniapedrazzi (changes requested):
I didn't had time to test the examples yet, but here are some initial comments.
I'm also trying to fix the test suite on the master/develop branches so that we could then run the tests in this branch.

katerynaCh:
> I didn't had time to test the examples yet, but here are some initial comments. I'm also trying to fix the test suite on the master/develop branches so that we could then run the tests in this branch.

Thank you! I have incorporated the changes.

stefaniapedrazzi (dismissed):
Thank you!

tsampazk:
Hey @katerynaCh, thank you for the new tool. I will do a thorough review soon. Taking a quick look, i noticed that doc entries for the ROS1/2 nodes seem to be missing. In the meantime, could you please add entries in the index lists [here (ros1)](https://github.com/opendr-eu/opendr/blob/gesture/projects/opendr_ws/README.md#rgb-input) and [here (ros2)](https://github.com/opendr-eu/opendr/blob/gesture/projects/opendr_ws_2/README.md#rgb-input) that link to corresponding sections [here](https://github.com/opendr-eu/opendr/blob/gesture/projects/opendr_ws/src/opendr_perception/README.md) and [here](https://github.com/opendr-eu/opendr/blob/gesture/projects/opendr_ws_2/src/opendr_perception/README.md)?

Let me know if you have any issues i can assist with!

tsampazk (approved):
Thank you @katerynaCh! I went ahead and applied various minor fixes that i will list below.

1. Tool webcam demo works really well, just fixed the max-hands argument type that was passed as a str and through an error when trying to change it via command line
2. Added some checks in the learner download method so the model doesn't get redownloaded if it already exists
3. Renamed the demo to webcam_demo to match most common naming in other tools
4. ROS1 node worked out of the box, just renamed the topics to match naming in other nodes (modified the doc appropriately too)
5. ROS2 node had an issue in the bridge conversion and threw an error which i fixed with a try/except because i am almost certain that if i straight up applied the float conversion, other ROS nodes will break, so that takes care of it.
6. Similar to ROS1 i modified the topic names to match other nodes (in the doc too) and changed the run command to the correct one (also noticed that some other tools had the same issue in the run command and fixed it).

All-in-all everything else looks good to me, thanks!

katerynaCh:
Thank you @tsampazk @stefaniapedrazzi !

PR #431, opened 2023-05-02
Debug test failures on master branch

stefaniapedrazzi:
* Disable mobile manipulation test
* PyTorch version bump
* Fixed ONNX and loading interface changes
* Fixed torchvision arguments
* Removed inplace operation from edgespeechnets
* Wheel version downgrade due to bug in the recent version
* Added missing dependencies
* Updated DCN loading to be backward compatible
* Update macos testing evironment from deprecated `macos-10.15` to `macos-11`

PR #439, opened 2023-06-05
Update PyTorch version in install_nvidia.sh

stefaniapedrazzi:
In  #431 we updated the module versions but this script was not updated:
* PyTorch from 1.9.0 to 1.13.1
* Torchvision from 0.10.0 to 0.14.1

In this PR, I propose some changes to update the Python modules version but I'm not sure if they will work on the NVIDIA TX2/AGX/NX .

### DIFF
--- a/bin/install_nvidia.sh
+++ b/bin/install_nvidia.sh
@@ -72,19 +72,19 @@ pip3 install configparser
 
 # Install Torch
 sudo apt-get install --yes libopenblas-dev cmake ninja-build
-TORCH=torch-1.9.0-cp36-cp36m-linux_aarch64.whl
-wget https://nvidia.box.com/shared/static/h1z9sw4bb1ybi0rm3tu8qdj8hs05ljbm.whl -O torch-1.9.0-cp36-cp36m-linux_aarch64.whl
+TORCH=torch-1.13.0a0+d0d6b1f2.nv22.10-cp38-cp38-linux_aarch64.whl
+wget https://developer.download.nvidia.com/compute/redist/jp/v502/pytorch/$TORCH -O $TORCH
 
 pip3 install Cython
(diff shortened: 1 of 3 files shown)


Pavlos-Tosidis:
Hello Stefania, I am gonna have a look into that and update to the closest version available soon

omichel:
Hi @Pavlos-Tosidis, did you have a chance to look into this?

Pavlos-Tosidis:
Yeah, meant to comment on it on todays TMT meeting.
So, 1.9 is the latest version supported by TX2. 
Unfortunately, even Nvidia stopped maintaining the JetPack for TX2 so we have to keep the current one (Ubuntu 18.04, with everything that this implies, e.g. python3.6.9).  But since the installation on the embedded devices is not the same as the toolkit for Desktop, updating the repo on the docker images to pull the changes made to the master branch and adding a couple of new dependencies worked fine. (At least all scripts under /projects/python). 
The question now is if we want to keep supporting TX2 and keep the current images and just update the repo used whenever we add/update something, or ditch TX2 or make a new image for NX devices, which support newer JetPack versions and update all dependencies to be the same, if applicable, with the Desktop version.

omichel:
In that case, I believe we should drop support for TX2 in the main branch of this repository otherwise we will be blocked with updating PyTorch. Using willing to use the unsupported TX2 may still revert to an older commit on this repo or older images (which we should document).

passalis:
One other possibility would be to keep the installation for tx-2 only for the tools not failing, i.e., keep the script as it is and add a notice with a list of tools that do not work in this platform (unless nothing works in the updated version, so this should be indeed tackled in a different manner).

passalis:
We can keep the current version on TX-2 image, so we can close this for now.

PR #463, opened 2023-09-18
GPU installation fix

tsampazk:
Fixes #461 

This PR fixes GPU detectron2 and torch installation in install.sh and introduces a temporary fix by removing problematic tools (Continual SLAM and Panoptic Segmentation). Detectron2 installation taken from [here](https://detectron2.readthedocs.io/en/latest/tutorials/install.html) (CUDA 11.3) and PyTorch installation taken from [here](https://pytorch.org/get-started/previous-versions/) (v1.13.1 -> Linux and Windows -> CUDA 11.6).

GPU installation was tested on `develop` branch with:
- NVIDIA drivers 460.106.00:
  1. Software & Updates 
  2. Additional Drivers
- CUDA 11.2 (required for mxnet): 
  1. cuda_11.2.0_460.27.04_linux.run from [here](https://developer.nvidia.com/cuda-11.2.0-download-archive?target_os=Linux&target_arch=x86_64&target_distro=Ubuntu&target_version=2004&target_type=runfilelocal)
  2. Linux -> x86_64 -> Ubuntu-> 20.04 -> runfile local  
  3. Skip driver installation
- cuDNN installed through [here](https://developer.nvidia.com/rdp/cudnn-archive)
  1. Download cuDNN v8.1.0 (January 26th, 2021), for CUDA 11.0,11.1 and 11.2
  2. cuDNN Runtime Library for Ubuntu20.04 x86_64 (Deb)

Installation of Panoptic Segmentation and Continual SLAM fail during building probably due to mismatches between CUDA versions (system 11.2 vs torch 11.6), so install.sh entirely removes their src directories.

Detectron2 installs fine but seems to fail the [tests](https://github.com/opendr-eu/opendr/actions/runs/6220543485/job/16880788204?pr=463):

<details>
  <summary>Spoiler warning</summary>

ERROR: test_single_demo_grasp (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: test_single_demo_grasp
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/unittest/loader.py", line 436, in _find_test_path
    module = self._get_module_from_name(name)
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
[...]

PR #466, opened 2023-09-25
Minor fix on yolov5 webcam demo

tsampazk:
When running yolov5 webcam demo, i ran into this error:

```
RuntimeError: cuDNN error: CUDNN_STATUS_NOT_INITIALIZED
```

when running `infer`.

This is fixed by importing torch (even though it is not directly used) which seems to initialize it properly, similar to inference demo which works fine.

This issue seems to originate from the latest gpu installation of the toolkit as described in #463.

### DIFF
--- a/.github/workflows/tests_suite.yml
+++ b/.github/workflows/tests_suite.yml
@@ -14,6 +14,10 @@ defaults:
   run:
     shell: bash
 
+env:
+    # XXX: Workaround for https://github.com/actions/cache/issues/1141
+    SEGMENT_DOWNLOAD_TIMEOUT_MINS: 5
+
 jobs:
   test-sources:
     if: ${{ contains(github.event.pull_request.labels.*.name, 'test sources') || github.event_name == 'schedule' }}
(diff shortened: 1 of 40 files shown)


omichel (approved):
Thank you.

tsampazk:
Hmm it seems that our tests ignore `# NOQA`. As far as i can tell this originates from `pyflakes` which doesn't seem to support `noqa`. I will look into some dummy usage of torch as a workaround.

omichel:
Do we really want to print this message to the console? I believe we could implement a more silent dummy usage, e.g., assign this version value to a dummy variable instead of printing it in the console. Also, it would be nice to add a comment here to explain why we need this dummy assignment.

tsampazk:
> Do we really want to print this message to the console? I believe we could implement a more silent dummy usage, e.g., assign this version value to a dummy variable instead of printing it in the console. Also, it would be nice to add a comment here to explain why we need this dummy assignment.

Yeah, i think that assigning and not using will still fail the pep8 tests for variable assigned but not used. On the other hand silencing print is too much trouble because one would need to manipulate stdout. I entirely removed print which didn't fail the pyflakes test locally (my IDE still complains), let's see if the CI is ok with it.

tsampazk:
It seems it passed the sources tests fine :smile:. Sorry for the extra trouble Olivier.

omichel (approved):
No problem. That's fine.
Good to see it passed the tests.

passalis (approved):
Thank you!

PR #467, opened 2023-09-26
New utility feature class filtering for Object Detectors 2D

### LINKED ISSUE #464: Object detectors 2D class filtering
Create a learner wrapper class to filter the detection output of an object detector's infer method. This class should be initialized providing a valid subset of specific classes that the user wants to detect.

katdimitris:
Implemented class filtering for object detectors using a wrapper. The wrapper accepts a Learner object and a list of desired classes for detection. These classes should be within the set that the learner is capable of detecting. An inference demo for YOLOv5 is also available.

Closes #464.

### DIFF
--- a/src/opendr/perception/object_detection_2d/utils/class_filter_wrapper.py
+++ b/src/opendr/perception/object_detection_2d/utils/class_filter_wrapper.py
@@ -57,7 +57,7 @@ class FilteredLearnerWrapper:
               nms_thresh=None, nms_topk=None, post_nms=None, extract_maps=None):
 
         # match variable names
-        if isinstance(self.learner, NanodetLearner):
+        if isinstance(self.learner, NanodetLearner) and input is not None:
             img = input
 
         if img is None:
--- a/src/opendr/perception/object_detection_2d/utils/class_filter_wrapper.py
+++ b/src/opendr/perception/object_detection_2d/utils/class_filter_wrapper.py
@@ -1,4 +1,12 @@
 from opendr.engine.target import BoundingBox, BoundingBoxList
+from opendr.perception.object_detection_2d.centernet.centernet_learner import CenterNetDetectorLearner
+from opendr.perception.object_detection_2d.detr.detr_learner import DetrLearner
+from opendr.perception.object_detection_2d.gem.gem_learner import GemLearner
+from opendr.perception.object_detection_2d.retinaface.retinaface_learner import RetinaFaceLearner
+from opendr.perception.object_detection_2d.ssd.ssd_learner import SingleShotDetectorLearner
+from opendr.perception.object_detection_2d.yolov3.yolov3_learner import YOLOv3DetectorLearner
+from opendr.perception.object_detection_2d.yolov5.yolov5_learner import YOLOv5DetectorLearner
+from opendr.perception.object_detection_2d.nanodet.nanodet_learner import NanodetLearner
 
 
 class FilteredLearnerWrapper:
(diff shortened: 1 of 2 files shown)


src/opendr/perception/object_detection_2d/utils/class_filter_wrapper.py
```diff
@@ -0,0 +1,137 @@
...
+            raise ValueError(
+                f"The following classes are not detected by this detector: {', '.join(invalid_classes)}")
+
+    def infer(self, img=None, threshold=None, keep_size=None, input=None, conf_threshold=None, iou_threshold=None,
+              nms_max_num=None, nms_threshold=None, scales=None, mask_thresh=None, size=None, custom_nms=None,
+              nms_thresh=None, nms_topk=None, post_nms=None, extract_maps=None):
+
+        # match variable names
+        if isinstance(self.learner, NanodetLearner):
+            img = input
```
tsampazk, line 62: For the wrapper to serve as a drop in replacement, we should assume that infer will be used without a keyword argument, e.g. in the nanodet webcam demo:

`boxes = filtered_nanodet.infer(img, conf_threshold=0.35, iou_threshold=0.6, nms_max_num=20)`

This leads to `input=None` which is then assigned to img. I suggest to add another check:

```suggestion
        if isinstance(self.learner, NanodetLearner) and input is not None:
            img = input
```

And change the infer to use `img` instead (next suggestion).

src/opendr/perception/object_detection_2d/utils/class_filter_wrapper.py
```diff
@@ -0,0 +1,137 @@
...
+            boxes = self.learner.infer(img)
+
+        elif isinstance(self.learner, NanodetLearner):
+            if conf_threshold is None:
+                conf_threshold = 0.35
+            if iou_threshold is None:
+                iou_threshold = 0.6
+            if nms_max_num is None:
+                nms_max_num = 100
+            boxes = self.learner.infer(input, conf_threshold, iou_threshold, nms_max_num)
```
tsampazk, line 97: If `input` is provided it will be assigned to `img` anyway so we can use `img` here instead.

```suggestion
            boxes = self.learner.infer(img, conf_threshold, iou_threshold, nms_max_num)
```

src/opendr/perception/object_detection_2d/utils/class_filter_wrapper.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+class FilteredLearnerWrapper:
+    def __init__(self, learner, allowed_classes=None):
+        self.learner = learner
+        self.allowed_classes = allowed_classes if allowed_classes is not None else []
+
+        if isinstance(self.learner,
+                      (CenterNetDetectorLearner, YOLOv3DetectorLearner, YOLOv5DetectorLearner, NanodetLearner,
+                       RetinaFaceLearner, SingleShotDetectorLearner)):
```
tsampazk, line 33: RetinaFace basically detects only faces so i am not sure whether it can be useful to filter its output. Similar to that, SSD is only a person detector.

I suggest we keep support for SSD since one may train it to detect some other class too, but drop support for RetinaFace since it only serves to detect faces. Any opinions are welcome.

tsampazk (changes requested):
Thank you @katdimitris, very nice job. I have added some comments. Mainly handling nanodet img/input is a bit problematic.

tsampazk (approved):
Looks good to me, thank you @katdimitris!

passalis (approved):
Thank you!

PR #468, opened 2023-09-28
[New tool] YOLOv8 Segmentation

tsampazk:
This PR adds a new tool using ultralytics YOLOv8 for segmentation usage.

WIP

### DIFF
--- a/docs/reference/yolov8-seg.md
+++ b/docs/reference/yolov8-seg.md
@@ -1,0 +1,81 @@
+## YOLOv5DetectorLearner module
+
+The *yolov5* module contains the *YOLOv5DetectorLearner* class, which inherits from the abstract class *Learner*.
+
+### Class YOLOv5DetectorLearner
+Bases: `engine.learners.Learner`
+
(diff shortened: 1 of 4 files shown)


tsampazk:
Since this tool is not a main tool, but just nice-to-have, we can go ahead and review and merge it. ROS1/2 nodes can be added later if needed. @thomaspeyrucain do you think it can prove useful? It provides bounding boxes and segmentation masks for the objects detected within, with very good performance in terms of both speed and accuracy. The classes that can be detected are listed [here](https://github.com/ultralytics/ultralytics/blob/9aaa5d5ed0e5a0c1f053069dd73f12b845c4f282/ultralytics/cfg/datasets/coco.yaml#L17) with the exception that within OpenDR, class 0 is background and class 80 is person.

thomaspeyrucain:
Hello @tsampazk ,
I think it would be really useful and can make us able to grasp more types of objects more effectively

tsampazk:
> Hello @tsampazk , I think it would be really useful and can make us able to grasp more types of objects more effectively

Alright, happy to hear that, i will add the ROS nodes in a subsequent PR.

PR #469, opened 2023-09-29
Apply cuDNN init fix to all Object Detectors 2D

tsampazk:
It seems that the fix applied to yolov5 webcam demo for the cudnn not initialized error (#466), needs to be applied to all object detectors 2D, as i encountered it when running other demos too.

I removed the fix from the webcam demo and added it to the `object_detectors_2d` module `__init__.py` file which seems to solve it for all detectors.


passalis (approved):
Thank you!

omichel (approved):
That makes sense.

PR #471, opened 2023-10-04
Updated test_suite_develop.yml based on latest test_suite.yml

tsampazk:
test_suite_develop.yml should be identical with test_suite.yml to run scheduled tests on `develop` as stated [here](https://github.com/opendr-eu/opendr/blob/develop/.github/workflows/tests_suite_develop.yml#L3-L4).


passalis (approved):

omichel (approved):
Yes, that's correct.

PR #472, opened 2023-10-19
Fix fmpgmapping

joseab10:
Fixed a bug when publishing the full posterior map where the comparison for the map model was done with the wrong enums, leading to publishing the wrong initial alpha and beta parameters for uninformed priors.


omichel (approved):
That looks good to me.
Thank you.

passalis (approved):
Thank you!

PR #478, opened 2023-11-13
Merge `develop` into `master`

passalis:
This PR aims to synchronize recent `develop` changes into `master`. 

This PR also bumps `av` version from `8.0.1` to `8.0.3` to resolve installation failure, as well as a minor fix to HR pose estimation which didn't download all necessary files.

PR #481, opened 2023-11-13
Installation fix for av

tsampazk:
Currently added `libavdevice-dev` and `libavfilter-dev` on main system installation to fix initial error.


tsampazk:
Issue seems to be fixed by changes in #478.

PR #483, opened 2023-11-14
Changed python dependencies installation

tsampazk:
Changed command that parses the created python prerequisites/dependencies files during install.sh so that it installs the pip packages one-by-one. This allows the install script to continue running even after one pip installation fails.

This is not tested thoroughly so i will look into it some more.

~This PR is intended to be merged to master after https://github.com/opendr-eu/opendr/pull/478.~

Update: Case in point, on commit 400b4165c9e66bf45f96f4eb829c3c65b802ed0a , the installation of one package (igibson) failed, but the rest of the installation went smoothly, causing the python env to be created normally for other tools and all the tests to run properly apart from multi-object-search which uses the package that failed.

However, this failure is new and most probably related to this new package installation procedure. More testing is required.

Update: Attempted to filter the temporary python dependencies file by removing duplicates and keeping only strict (==version) versions of packages. Getting this conflict:

`ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
scikit-image 0.21.0 requires imageio>=2.27, but you have imageio 2.6.0 which is incompatible.`

---------

Update: Reworked dependency file filtering from the ground up that can act as a dependency debugging tool. Also slightly cleared up some existing dependencies.

**The new methods warn for possible dependency constraint clashes and throw an error when a definite clash occurs. This happens before the actual pip installations which can help diagnose first-level toolkit dependency clashes early.**

### DIFF
--- a/dependencies/install.sh
+++ b/dependencies/install.sh
@@ -20,11 +20,11 @@ if [ -f "linux_dependencies.txt" ]; then
        rm linux_dependencies.txt
 fi
 if [ -f "python_prerequisites.txt" ]; then
-       pip install -r python_prerequisites.txt
+       cat python_prerequisites.txt | while read PACKAGE; do pip install $PACKAGE; done
        rm python_prerequisites.txt
 fi
 if [ -f "python_dependencies.txt" ]; then
-       pip install -r python_dependencies.txt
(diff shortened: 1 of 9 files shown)


passalis (approved):
Thanks!!!

omichel (approved):
Thank you.

PR #486, opened 2023-11-14
Prepare v3.0.0 Release

passalis:
This aims to prepare v3.0.0 release.


passalis:
Only nine failures, that actually sounds very promising :laughing:

passalis:
This is now stable. We wait for all features to be included in v3.0.0 to be merged and then we can continue from there.

passalis:
Let's see if this still works...

passalis:
The failure seems to be related to the artifacts being deleted before GEM test run. Not sure why this happened, but now it seems to work. Let's wait for develop to be merged and then see if this happens again...

omichel (dismissed):
That looks all good to me.
We just need to set the release date in the CHANGELOG.md.

PR #476, opened 2023-11-07
Finetuned yolov5 for trucks

BillMousta:
This pull request introduces a new download method within the YOLOv5DetectorLearner class, enabling automated fetching of pre-trained models directly into the local environment. Additionally, it showcases an inference demo leveraging a YOLOv5s model fine-tuned for enhanced truck detection accuracy. This demonstration serves as a practical example of the model's improved performance in identifying trucks.

PR #487, opened 2023-11-15
Yolov5 trucks fixes

tsampazk:
Applied the fixes from the late review on #476.

@BillMousta 

One general comment: I am not sure about the naming of the new feature. We use the word trucks which is not really accurate since the model is trained on tractors, so it could be misleading. Reading trucks i would intuitively expect pickup trucks or semis. Let me know if i misunderstand something regarding the naming. If you agree i can go ahead and rename everything in this PR.

### DIFF
--- a/docs/reference/object-detection-2d-yolov5.md
+++ b/docs/reference/object-detection-2d-yolov5.md
@@ -76,7 +76,7 @@ Parameters:
   If True, enables maximum verbosity.
 - **url**: *str, default=OpenDR FTP URL*\
   URL of the FTP server.
-- **model_name**: name of model ftp server, *default = 'yolov5s_finetuned_in_trucks.pt'.*\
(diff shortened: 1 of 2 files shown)


BillMousta:
> Applied the fixes from the late review on #476.
> 
> @BillMousta
> 
> One general comment: I am not sure about the naming of the new feature. We use the word trucks which is not really accurate since the model is trained on tractors, so it could be misleading. Reading trucks i would intuitively expect pickup trucks or semis. Let me know if i misunderstand something regarding the naming. If you agree i can go ahead and rename everything in this PR.

The primary objective of this fine-tuned model was to enhance the vehicle detection capabilities, particularly in the agricultural sector. We operated under the assumption that trucks and tractors share similar visual characteristics. Therefore, during training, we categorized both trucks and tractors under the 'trucks' class. This approach resulted in a more robust model, compared to the alternative of introducing a separate 'tractors' class.

passalis (approved):
Thanks!

omichel (approved):

